PR #15, opened 2018-02-21
Add ESLint, fix files

BafS:


lib/key/check.js
```diff
@@ -81,4 +81,4 @@ function keyCheck(info, email, expectEncrypted) {
     return info;
 }
 
-module.exports = keyCheck;
\ No newline at end of file
+export default keyCheck;
```
dhoko: why ES6 module ?

lib/key/decrypt.js
```diff
@@ -3,54 +3,53 @@ const { getKeys, pickPrivate } = require('./utils');
...
             }
-            return Promise.reject(err);
-        }
-    });
+        });
 }
 
-module.exports = { decryptPrivateKey, decryptSessionKey };
\ No newline at end of file
+export default { decryptPrivateKey, decryptSessionKey };
```
dhoko: if ES6 module, use export not the default

package.json
```diff
@@ -31,6 +32,9 @@
     "babel-preset-es2015": "^6.24.1",
     "babel-preset-stage-0": "^6.24.1",
     "babelify": "^7.3.0",
-    "browserify": "^14.4.0"
```
dhoko: Did you try to build ?

dhoko:
We will add ES6 module support later via the bridge https://www.npmjs.com/package/@std/esm for the API. But it will be another PR.

PR #20, opened 2018-02-23
Bootstrap tests

BafS:


KAYLukas:
The utilities functions tested here now mostly test the openpgpjs utilities helpers as we simply use their utility functions.

PR #26, opened 2018-03-29
Export the commonjs module for the web

dhoko:
:warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: 

We must RFR the lib,  this PR is only about doing a Quick and dirty patch.

:warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: :warning: 

We should improve the API, as it's done the API is poor. A huge object, we should export modules.

2018 and there is a `require('es6-promise').Promise;` NOPE.

```js
global.btoa = require('btoa');
global.atob = require('atob');
global.Promise = require('es6-promise').Promise;
global.openpgp = require('openpgp');
```

This is not because making global is easy and magic we should do it. We should not.
When you need openpgp do the require inside the module not as it's done today.

AS you need to set some keys, 

```js
openpgp.config.integrity_protect = true;
openpgp.config.use_native = true;
```

You can create a custom wrapper.
You should load asmCrypto via this module. Or not at all (_best solution as you don't need it and no tree shacking_).

Inside package.json
```
  "main": "index.js",
  "browser": "browser/lib.js",
```

It means, íf you require from node.js the lib it will use `index.js`. We use the file from the key browser is only when we create a bundle (commonjs) inside a front-end. (ex external app). 

As it is this PR works. But it's bad.


dhoko:
PS we need to test the lib.

BafS:
For the test, a PR is waiting https://github.com/ProtonMail/pmcrypto/pull/20 :trollface:

bartbutler:
I think the Promise global can be completely removed.

KAYLukas:
We are not using browserify at this point.

Regarding
```
global.btoa = require('btoa');
global.atob = require('atob');
global.Promise = require('es6-promise').Promise;
global.openpgp = require('openpgp');
```
This is now only included in the node version.

PR #35, opened 2018-10-19
Add subkey keyIDs in info()

valentinbonneaud:
Add a `subKeysFingerprints` element in the `info` function. This element contains the fingerprint of the subkeys for the given PGP key.

### DIFF
--- a/dist/pmcrypto-node.js
+++ b/dist/pmcrypto-node.js
@@ -26723,8 +26723,8 @@ const packetInfo = (packet, key) => {
     return createPacketInfo(packet, key);
 };
 
-const getSubkeysFingerprints = (key) => {
-    return key.subKeys ? key.subKeys.map((subkey) => subkey.getFingerprint()) : [];
+const getSubkeysFingerprints = ({ subKeys = [] } = {}) => {
+    return subKeys.map((subkey) => subkey.getFingerprint());
 };
 
 const primaryUser = async (key, date) => {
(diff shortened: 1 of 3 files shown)


lib/key/info.js
```diff
@@ -27,6 +27,10 @@ const packetInfo = (packet, key) => {
     return createPacketInfo(packet, key);
 };
 
+const getSubkeysFingerprints = (key) => {
```
dhoko: You can remove the if via destructuring -> if key can't be null.
```suggestion
const getSubkeysFingerprints = ({ subKeys = [] } = {}) => {
```

valentinbonneaud: You're right, I updated the code. The key cannot be `null` at this point in the code. If it's the case, an error is thrown in the `getKeys` function.

dist/pmcrypto-node.js
```diff
@@ -26738,11 +26742,13 @@ async function keyInfo(rawKey, email, expectEncrypted = true, date = serverTime(
     const keys = await getKeys(rawKey);
 
     const algoInfo = keys[0].getAlgorithmInfo();
+    const mainFingerprint = keys[0].getFingerprint();
 
     const obj = {
         version: keys[0].primaryKey.version,
         publicKeyArmored: keys[0].toPublic().armor(),
-        fingerprint: keys[0].getFingerprint(),
+        fingerprint: mainFingerprint,
```
sanjanarajan: Could you add a FIXME so we deprecate this?

PR #37, opened 2018-10-23
Add date checks to pmcrypto

KAYLukas:


lib/key/check.js
```diff
@@ -74,3 +75,16 @@ export default function keyCheck(info, email, expectEncrypted) {
 
     return info;
 }
+
+export function dateChecks([ key ], date = serverTime()) {
+    const keys = [ key, ...key.subKeys ];
+    if (!keys.every(({ keyPacket }) => keyPacket.created < date)) {
```
dhoko: you can use some ;)

test/key/info.js
```diff
@@ -30,7 +30,47 @@ JLdxTBJgqQoLePyHnb4LEPZItOUOMiHRBLUV8PSj/dxLEmYlmbPFOTnE62izRX4Q
...
+RLfkk/sT5Ay2LtjvPUmynkGvI3LFRkkqiA4buDiFOV/pqdV6RGIXEOcyMyPJ
+hufD30wViByeDdxpJsFOlVnw7D02WDNUHomhN2m8ogLJyFbfrG8/c54VmX6O
+J5Tj5Die5J5mSWCHbi9O0TBBgCysDxAr7blqK2xBO85q+Kp0809lPiS97iDq
+uokpJQHZjIvfQ5/9tx1946Tvo0RX0A26JfOO+J68XA==
+=MPBF
+-----END PGP PUBLIC KEY BLOCK-----`;
+
+    test('creation test', async () => {
+        const { dateError } = await pmcrypto.keyInfo(creationkey);
+        assert.strictEqual(dateError, 'The self certifications are created with illegal times');
```
dhoko: Can you add more tests ?

KAYLukas: @sanjanarajan Do you know how to generate these keys?

sanjanarajan: What keys? You are already testing a key with invalid times, right?

sanjanarajan: That is sufficient

dhoko: There are 3 diff errors, we need tests for them.

sanjanarajan: Maybe you could just add a test for a normal key to make sure it doesn't throw

dhoko: There is always a test for this case, but it doesn't test what's missing for errors.

PR #39, opened 2018-11-11
Fix #38 - Upgrade to OpenPGP.js 4

### LINKED ISSUE #38: Upgrade to OpenPGP.js 4
See https://github.com/ProtonMail/Angular/issues/7862.

Another thing that I forgot to mention is that we increased the default number of S2K rounds, which makes things slower. Since we use bcrypt in addition to S2K for key derivation, it's probably safe to add `openpgp.config.s2k_iteration_count_byte = 96` to https://github.com/ProtonMail/pmcrypto/blob/master/lib/openpgp.js for performance

EpokK:
### Changes
- `openpgp.{message,key,signature,cleartext}.readArmored` and `openpgp.{message,key,signature}.read` are now asynchronous.

- `openpgp.encrypt` and `openpgp.sign` now take a `message` parameter instead of the `data`, `dataType` and `filename` parameters. You can use `pmcrypto.createMessage` or `pmcrypto.getCleartextMessage` to get a message object (previously, the default was a cleartext message for signing text and a normal message for signing binary data and encrypting text or binary).

- After calling `let keyring = new openpgp.Keyring()`, you now have to call `await keyring.load()` to read the keys.

### Functions impacted

* getKeys (async)
    * compressKey (async)
    * cloneKey (async)
* getMessage (async)
    * splitMessage (async)
* getSignature (async)
* getMatchingKey (async)
* encryptMessage (signature)
* signMessage (signature)


lib/message/utils.js
```diff
@@ -39,6 +39,10 @@ export function createMessage(source) {
 }
 
 export function signMessage(options) {
+    if (typeof options.data === 'string') {
+        options.message = createMessage(options.data);
```
twiss: You probably want `getCleartextMessage` here, since that was the default for signing strings. Also I think you'll need to handle `Uint8Array`s in both `encryptMessage` and `signMessage` as well (you can use `createMessage` for both).

EpokK: @twiss changes applied in this commit: 23fcc8b

lib/key/utils.js
```diff
@@ -67,21 +71,26 @@ export function getFingerprint(key) {
...
         loadPrivate: () => [],
         storePublic() {},
         storePrivate() {}
     });
 
+    await keyring.load();
+
     const keyids = signature.packets.map(({ issuerKeyId }) => issuerKeyId.toHex());
-    const [key = null] = keyids.map((keyid) => keyring.getKeysForId(keyid, true) || []).flatten().filter(Boolean);
+    const [key = null] = keyids
```
dhoko: why null ?

lib/key/utils.js
```diff
@@ -67,21 +71,26 @@ export function getFingerprint(key) {
...
+        .map((keyid) => keyring.getKeysForId(keyid, true) || [])
```
dhoko: instead of map + flatten + filter, use a reducer here ;)

twiss:
Looks good to me :+1: the last thing would be

> Another thing that I forgot to mention is that we increased the default number of S2K rounds, which makes things slower. Since we use bcrypt in addition to S2K for key derivation, it's probably safe to add `openpgp.config.s2k_iteration_count_byte = 96` to https://github.com/ProtonMail/pmcrypto/blob/master/lib/openpgp.js for performance

PR #42, opened 2018-12-05
Use "real" modules

mmso:
Since the client needs to select which bundle of openpgpjs to use, expect that it exists on the window or global object instead

Don't build anything with rollup.

Use esm instead for nodejs.
Use the native behavior for the browser.


lib/index.js
```diff
@@ -1,8 +1,2 @@
-import * as pmcrypto from './pmcrypto';
-
-/* START.BROWSER_ONLY */
-window.pmcrypto = pmcrypto;
-/* END.BROWSER_ONLY */
-
-export default pmcrypto;
-export * from './pmcrypto';
+require = require('esm')(module);
```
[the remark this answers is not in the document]
mmso: Hmm I was following https://www.npmjs.com/package/esm#getting-started this example here

I will double check but I think this is just hit when requiring this module from node.js

lib/index.js
```diff
@@ -1,8 +1,6 @@
-import * as pmcrypto from './pmcrypto';
+global.btoa = require('btoa');
+global.atob = require('btoa');
```
EpokK: Isn't `atob` here?
```suggestion
global.atob = require('atob');
```

lib/mail.js
```diff
@@ -0,0 +1,30 @@
...
+import { getMessage } from './message/utils';
+import { decryptMessage } from './message/decrypt';
+
+export async function checkMailboxPassword(prKey, prKeyPassCode, accessToken) {
+    if (typeof prKey === 'undefined') {
+        throw new Error('Missing private key.');
+    }
+
+    if (typeof prKeyPassCode === 'undefined') {
+        throw new Error('Missing Mailbox Password.');
```
EpokK:
```suggestion
        throw new Error('Missing mailbox password.');
```

lib/mail.js
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    const message = await getMessage(accessToken);
+
+    try {
+        // this is the private key, use this and decryptMessage to get the access token
+        const { data } = await decryptMessage({ message, privateKeys: [privateKey] });
+        return { password: prKeyPassCode, token: data };
+    } catch (err) {
+        throw new Error('Wrong Mailbox Password.');
```
EpokK:
```suggestion
        throw new Error('Wrong mailbox password.');
```

lib/message/utils.js
```diff
@@ -107,22 +106,24 @@ export async function handleVerificationResult({ data, filename = 'msg.txt', sig
...
-        return verificationResults.filter(({ verified }) => verified === SIGNED_AND_VALID).reduceRight(
-            (acc, result) => {
-                if (acc.verified !== SIGNED_AND_VALID) {
-                    acc.verified = result.verified;
-                    acc.data = arrayToBinaryString(result.data);
+        return verificationResults
+            .filter(({ verified }) => verified === SIGNED_AND_VALID)
+            .reduceRight(
+                (acc, result) => {
+                    if (acc.verified !== SIGNED_AND_VALID) {
```
EpokK: Is it really possible to have `acc.verified !== SIGNED_AND_VALID` if we just filtered `verificationResults` like that `filter(({ verified }) => verified === SIGNED_AND_VALID)`?

lib/message/utils.js
```diff
@@ -107,22 +106,24 @@ export async function handleVerificationResult({ data, filename = 'msg.txt', sig
         });
         const verificationResults = await Promise.all(verificationPromises);
 
-        return verificationResults.filter(({ verified }) => verified === SIGNED_AND_VALID).reduceRight(
-            (acc, result) => {
-                if (acc.verified !== SIGNED_AND_VALID) {
-                    acc.verified = result.verified;
-                    acc.data = arrayToBinaryString(result.data);
+        return verificationResults
+            .filter(({ verified }) => verified === SIGNED_AND_VALID)
```
EpokK: Maybe we should combine `reduceRight` and `filter`?

lib/message/utils.js
mmso: Yes you are right

lib/mail.js
```diff
@@ -0,0 +1,30 @@
+import { decryptPrivateKey } from './key/decrypt';
+import { getMessage } from './message/utils';
+import { decryptMessage } from './message/decrypt';
+
+export async function checkMailboxPassword(prKey, prKeyPassCode, accessToken) {
```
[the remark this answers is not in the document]
mmso: Yes I agree, lets make it in another iteration

lib/index.js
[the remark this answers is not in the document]
mmso: Yes it's just to make it requireable from node.js, in our apps it will use the real module

lib/openpgp.js
```diff
@@ -1,23 +0,0 @@
...
-/**
- * @link https://github.com/vibornoff/asmcrypto.js/issues/121
- */
-asmCrypto.random.skipSystemRNGWarning = true;
-
-const { hardwareConcurrency = 1 } = window.navigator || {};
-openpgp.initWorker({ path: 'openpgp.worker.min.js', n: hardwareConcurrency });
-/* END.BROWSER_ONLY */
-
-openpgp.config.integrity_protect = true;
```
KAYLukas: I don't see a reason to move this to the app itself. I think we should have a global setup controlled in this app. That would allow us to handle defaults that are changing in the main library more easily.

mmso: Agreed :)

After discussion with @KAYLukas and @sanjanarajan we decided to leave the configuration in pmcrypto but the worker initialization inside the user of the library because the path to the worker will change depending on the browser, and nodejs does not support workers, and maybe (I can't give a concrete example) we don't want to have the worker in some app or in some browser. This just lets us be more flexible.

dhoko: We need to use a function to pass the path to pmcrypto, no ?

mmso: I mean I'm fine to add a helper function to create a worker in openpgp to reduce the duplication of two lines, but I think it should not automatically create one.

pmcrypto is just a library of helper functions, it should not do magic in my opinion.

PR #47, opened 2019-01-25
Feat/tunnel

wussler:
New functions for ECDHE exchange


lib/key/utils.js
```diff
@@ -101,3 +104,29 @@ export async function cloneKey(inputKey) {
...
+    return openpgp.crypto.publicKey.elliptic.ecdh.genPublicEphemeralKey(
+        oid,
+        curveObj.cipher,
+        curveObj.hash,
+        Q,
+        fingerprint
+    );
+}
+
+export async function genPrivateEphemeralKey(curve, V, d, fingerprint) {
```
dhoko: Can you add the JSDoc ?
Instead of having 4 args, can you use destructuring ?

lib/key/utils.js
```diff
@@ -101,3 +104,29 @@ export async function cloneKey(inputKey) {
     const [key] = await getKeys(inputKey.toPacketlist().write());
     return key;
 }
+
+// ECDHE
+export async function genPublicEphemeralKey(curve, Q, fingerprint) {
```
dhoko: Can you add the JSDoc ?

lib/key/utils.js
```diff
@@ -2,11 +2,14 @@ import { openpgp } from '../openpgp';
...
 // returns promise for generated RSA public and encrypted private keys
-export const generateKey = (options) => {
+export async function generateKey(options) {
     options.date = typeof options.date === 'undefined' ? serverTime() : options.date;
     return openpgp.generateKey(options);
-};
-export const generateSessionKey = (algorithm) => openpgp.crypto.generateSessionKey(algorithm);
+}
+
+export async function generateSessionKey(algorithm) {
```
dhoko: `export const generateSessionKey = openpgp.crypto.generateSessionKey`
No need to wrap it again ;) and it's already async

wussler: Thanks, I'll remove from the commit

PR #49, opened 2019-02-13
Feat/tunnelv5

wussler:


lib/key/utils.js
```diff
@@ -2,11 +2,14 @@ import { openpgp } from '../openpgp';
...
 // returns promise for generated RSA public and encrypted private keys
-export const generateKey = (options) => {
+export async function generateKey(options) {
     options.date = typeof options.date === 'undefined' ? serverTime() : options.date;
     return openpgp.generateKey(options);
-};
-export const generateSessionKey = (algorithm) => openpgp.crypto.generateSessionKey(algorithm);
+}
+
+export async function generateSessionKey(algorithm) {
```
dhoko:
```suggestion
export const generateSessionKey = openpgp.crypto.generateSessionKey;
```

lib/key/utils.js
```diff
@@ -101,3 +104,67 @@ export async function cloneKey(inputKey) {
...
+ *
+ * @param {Array<Options>}         Private key d, public part of ECDHE V, fingerprint and curve (name or OID)
+ * @returns {Promise<Uint8Array>}  Generated ephemeral secret
+ * @async
+ */
+export async function genPrivateEphemeralKey(options) {
+    const curveObj = new openpgp.crypto.publicKey.elliptic.Curve(options.Curve);
+    const oid = new openpgp.OID(curveObj.oid);
+    const S = await openpgp.crypto.publicKey.elliptic.ecdh.genPrivateEphemeralKey(curveObj, options.V, options.d);
+    const cipher_algo = openpgp.enums.read(openpgp.enums.symmetric, curveObj.cipher);
```
dhoko: Can you change the format of var to `cipherAlgo` ;)

lib/key/utils.js
```diff
@@ -101,3 +104,67 @@ export async function cloneKey(inputKey) {
...
+        false
+    );
+
+    return {V, Z}
+}
+
+/**
+ * Generate ECDHE secret from private key and public part of ephemeral key
+ *
+ * @param {Array<Options>}         Private key d, public part of ECDHE V, fingerprint and curve (name or OID)
```
dhoko: Wrong type, it's an Object.
Can you use destructuring to improve the readability of options ? (_it's easier to doc as we can do per options + types + what we need without looking at the body_)

lib/key/utils.js
```diff
@@ -101,3 +104,67 @@ export async function cloneKey(inputKey) {
...
+ * Generate ECDHE key and secret from public key
+ *
+ * @param {Array<Options>}                              Public key Q, fingerprint and curve (name or OID)
+ * @returns {Promise<{V: Uint8Array, Z: Uint8Array}>}   Returns public part of ephemeral key and generated ephemeral secret
+ * @async
+ */
+export async function genPublicEphemeralKey(options) {
+    const curveObj = new openpgp.crypto.publicKey.elliptic.Curve(options.Curve);
+    const oid = new openpgp.OID(curveObj.oid);
+    const { V, S } = await openpgp.crypto.publicKey.elliptic.ecdh.genPublicEphemeralKey(curveObj, options.Q);
```
dhoko: No need to wrap it with try/catch ? Is it expected ?

lib/key/utils.js
```diff
@@ -101,3 +104,67 @@ export async function cloneKey(inputKey) {
...
 }
+
+/**
+ * Generate ECDHE key and secret from public key
+ *
+ * @param {Array<Options>}                              Public key Q, fingerprint and curve (name or OID)
+ * @returns {Promise<{V: Uint8Array, Z: Uint8Array}>}   Returns public part of ephemeral key and generated ephemeral secret
+ * @async
+ */
+export async function genPublicEphemeralKey(options) {
```
dhoko: idem about destructuring (cf next comment)

dhoko:
Can you add tests too ?

lib/key/utils.js
wussler: It is try/catched upstream in the server

wussler:
> Can you add tests too ?

Tests are done in Openpgpjs and Slim-API, I will add them soon

PR #51, opened 2019-02-19
Add pm-srp required exports

mmso:
Adds:
* SHA512 hash, exposed as `SHA512`
* MD5 hash, exposed as `unsafeMD5`
* `arrayToHexString`
* `createCleartextMessage`

Changes:
* `getCleartextMessage` now reads an armored message instead of creating one. To create one the `createCleartextMessage` should be used. This is to have consistency with the `getMessage`, `createMessage`, `getSignature` etc functions. It does not seem to be used in the encryption server, it is used in the Angular project so it needs to be updated there.

Removes:
* `getHashedPassword` it's broken and not used in pmcrypto, encryption-server or angular.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -18,6 +18,7 @@ export const decodeUtf8Base64 = ifDefined((input) => decodeUtf8(decodeBase64(inp
 
 export const binaryStringToArray = (args) => openpgp.util.str_to_Uint8Array(args);
 export const arrayToBinaryString = (args) => openpgp.util.Uint8Array_to_str(args); // eslint-disable-line new-cap
+export const arrayToHexString = (args) => openpgp.util.Uint8Array_to_hex(args); // eslint-disable-line new-cap
 export const concatArrays = (args) => openpgp.util.concatUint8Array(args);
 
 export const sha512 = (args) => openpgp.crypto.hash.sha512(args);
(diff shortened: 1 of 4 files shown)


lib/utils.js
```diff
@@ -18,10 +18,11 @@ export const decodeUtf8Base64 = ifDefined((input) => decodeUtf8(decodeBase64(inp
 
 export const binaryStringToArray = (args) => openpgp.util.str_to_Uint8Array(args);
 export const arrayToBinaryString = (args) => openpgp.util.Uint8Array_to_str(args); // eslint-disable-line new-cap
+export const arrayToHexString = (args) => openpgp.util.Uint8Array_to_hex(args); // eslint-disable-line new-cap
 export const concatArrays = (args) => openpgp.util.concatUint8Array(args);
 
-export const getHashedPassword = (password) =>
-    localBtoa(arrayToBinaryString(openpgp.crypto.hash.sha512(binaryStringToArray(password))));
+export const sha512 = (args) => openpgp.crypto.hash.sha512(args);
+export const md5 = (args) => openpgp.crypto.hash.md5(args);
```
dhoko: Can you use aliases ?

mmso: No because of the way openpgp is dynamically set

dhoko: Can you explain ? I don't understand

mmso: In order for openpgp to become available for this library it needs to be set by the user of pmcrypto using `pmcrypto.init(openpgp)`. It's like that because the version of openpgp is dynamically loaded and some parts (requiring pmcrypto) of the application can be loaded before openpgp is loaded. So then using an alias would fail because it does not have openpgp yet.

PR #61, opened 2019-05-03
Add date option to reformatKey

mmso:
@twiss do you know if the user object should be more like

```
{ name:'Phil Zimmermann', email:'phil@openpgp.org' 
```

Because currently it's using the email for the name
```
{ name: email, email }
```

### DIFF
--- a/lib/key/utils.js
+++ b/lib/key/utils.js
@@ -11,9 +11,9 @@ export function generateSessionKey(algorithm) {
     return openpgp.crypto.generateSessionKey(algorithm);
 }
 
-export function reformatKey(privKey, email = '', passphrase = '') {
-    if (passphrase.length === 0) {
-        return Promise.reject(new Error('Missing private key passcode'));
+export function reformatKey({ privateKey, email = '', passphrase, date = serverTime() }) {
+    if (!passphrase) {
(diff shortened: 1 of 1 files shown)


twiss:
@mmso Normally yes, but ProtonMail uses the email as the name for the user id everywhere so it's fine for now. But ideally this function should probably just accept an options object similar to generateKey, and then handle it in the application.

mmso:
Good idea, I'll fix that

lib/key/utils.js
```diff
@@ -11,23 +11,9 @@ export function generateSessionKey(algorithm) {
     return openpgp.crypto.generateSessionKey(algorithm);
 }
 
-export function reformatKey(privKey, email = '', passphrase = '') {
-    if (passphrase.length === 0) {
-        return Promise.reject(new Error('Missing private key passcode'));
-    }
```
twiss: Maybe we want to keep this check? I.e. throw if `!options.passphrase`. The reason being that openpgp.js just returns an unencrypted key in that case, which might lead to a vulnerability. (The same thing goes for `generateKey`, really, but you might want to double check that we always pass a passphrase there before adding the check there.)

PR #69, opened 2020-01-22
Add typings

mmso:


index.d.ts
```diff
@@ -0,0 +1,138 @@
+declare module 'pmcrypto' {
```
vincaslt: When authoring definitions for libraries, you shouldn't declare it by name, it's by default going to be resolved correctly automatically. Since pmcrypt doesn't declare a global, you can just remote the `declare module 'pmcrypto' {` part and it should work fine.

cf https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-d-ts.html

vincaslt:
Nice work!

lib/index.d.ts
```diff
@@ -0,0 +1,188 @@
...
+    getEncryptedSubject: () => Promise<string>;
+    verify: () => Promise<number>;
+};
+
+export interface EncryptOptionsPmcrypto extends Omit<EncryptOptions, 'message'> {
+    data?: Uint8Array | string;
+    message?: message.Message;
+}
+
+export function encryptMessage(options: EncryptOptionsPmcrypto): Promise<EncryptResult>;
```
vincaslt:
```suggestion
export function encryptMessage(options: EncryptOptionsPmcrypto & { armor?: true, detached?: false }): Promise<EncryptResult>
export function encryptMessage(options: EncryptOptionsPmcrypto & { armor?: true, detached: true }): Promise<EncryptResult<string, undefined, string>>
export function encryptMessage(options: EncryptOptionsPmcrypto & { armor: false, detached?: false }): Promise<EncryptResult<undefined, message.Message>>
export function encryptMessage(options: EncryptOptionsPmcrypto & { armor: false, detached: true }): Promise<EncryptResult<undefined, message.Message, signature.Signature>>
export function encryptMessage(options: EncryptOptionsPmcrypto): Promise<EncryptResult<
    string | ReadableStream<String>,
    message.Message,
    string | ReadableStream<String> | signature.Signature
>>;
```

and:

```
export interface EncryptResult<D = string, M = undefined, S = undefined> {
    data: D;
    message: M;
    signature: S;
    sessionKey: SessionKey;
}
```

vincaslt: Based on openpgp typings

PR #73, opened 2020-02-27
catch signature verification errors produced by openpgp

econdepe:
In version 4.10, OpenPGP.js is rejecting some signature algorithms (most notably SHA1) with a custom message. We want to catch this message to display to the user in case an email with one of those invalid signatures is received.

### DIFF
--- a/lib/message/decrypt.js
+++ b/lib/message/decrypt.js
@@ -88,7 +88,7 @@ export async function decryptMessageLegacy(options) {
 /**
  * Decrypts the mime message and parses the body and attachments in the right structure.
  * @param options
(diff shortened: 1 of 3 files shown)


lib/message/decrypt.js
```diff
@@ -104,6 +105,7 @@ export async function decryptMIMEMessage(options) {
         getAttachments: () => Promise.resolve(attachments),
         getEncryptedSubject: () => Promise.resolve(encryptedSubject),
         verify: () => Promise.resolve(verified),
+        errors: () => Promise.resolve(errors),
```
mmso: Could be good to add this to the typescript definition

PR #78, opened 2020-03-25
Update decryptMessageLegacy response type

Swiip:
I couldn't makes old code typing match (`verified` property here https://github.com/ProtonMail/WebClient/blob/b2194e41d02cb7455f3228c73bd52573a24b0dff/src/app/message/factories/messageModel.js#L336), I discovered that `decryptMessageLegacy` was not good.

The result is not very elegant but it's what I end up looking at the code.


Swiip:
Outdated

PR #80, opened 2020-04-23
return signature when decrypting a message

econdepe:
It doesn't hurt, and can be useful for finding keys that sign the message


lib/index.d.ts
```diff
@@ -131,6 +131,7 @@ export function decryptSessionKey(options: {
 
 export interface DecryptResultPmcrypto extends DecryptResult {
     verified: VERIFICATION_STATUS;
+    signature: OpenPGPSignature;
 }
```
twiss, line 135: `decryptMessage` already returns a `signatures` property, each element of which has a `keyid` - this is already specified by `DecryptResult`.
It seems like `decryptMessageLegacy` doesn't conform to this though, it returns `signature: 0` which should be `verified: 0, signatures: []`.

lib/index.d.ts
```diff
@@ -151,6 +152,7 @@ export function decryptMIMEMessage(
     getEncryptedSubject: () => Promise<string>;
     verify: () => Promise<number>;
     errors: () => Promise<Error[] | undefined>;
+    getSignature: () => Promise<OpenPGPSignature>;
 };
```
twiss, line 156: And then personally, I would aim to make this return something similar to the `signatures` of `decryptMessage` - see https://github.com/DefinitelyTyped/DefinitelyTyped/blob/839738b973f370fa4d7e1fdad1fcb7deb528adf9/types/openpgp/index.d.ts#L5221-L5225.

lib/message/processMIME.js
```diff
@@ -55,7 +55,7 @@ const verifySignature = async ({ publicKeys, date }, data) => {
         signature
     });
 
-    return { subdata, verified };
+    return { subdata, verified, signature };
```
twiss, line 58: `verifyMessage` also returns a `signatures` array similar to `decryptMessage`. I would return that, instead.

lib/message/decrypt.js
```diff
@@ -93,7 +93,7 @@ export async function decryptMessageLegacy(options) {
 export async function decryptMIMEMessage(options) {
     const { data: rawData, verified: embeddedVerified, errors } = await decryptMessageLegacy(options);
 
-    const { body, mimetype, verified: pgpVerified, attachments, encryptedSubject } = await processMIME(
+    const { body, mimetype, verified: pgpVerified, attachments, encryptedSubject, signature } = await processMIME(
```
twiss, line 96: Since `decryptMessageLegacy` also return a signatures array, you should probably then concatenate the two arrays together, in this function. This corresponds to the fact that a PGP signature can appear both in the MIME structure and in the PGP message (cf. respectively `pgpVerified` and `embeddedVerified`, confusingly enough).
As written, you're only returning `signature` if it's in the MIME structure, which is the less common of the two.

lib/message/decrypt.js
```diff
@@ -91,13 +91,19 @@ export async function decryptMessageLegacy(options) {
...
     );
 
+    const {
+        body,
+        mimetype,
+        verified: pgpVerified,
+        attachments,
+        encryptedSubject,
+        signatures: mimeSignatures
+    } = await processMIME(options, rawData);
```
twiss, line 105: Your new names (`pgpSignatures` and `mimeSignatures`) are correct, but in combination with the previous ones (`embeddedVerified` and `pgpVerified`) it's very confusing.
I would propose `verified` and `signatures` for the first set, and `pgpMimeVerified` and `pgpMimeSignatures` for the second one.

lib/index.d.ts
```diff
@@ -129,7 +129,8 @@ export function decryptSessionKey(options: {
     passwords?: string | string[];
 }): Promise<SessionKey | undefined>;
 
-export interface DecryptResultPmcrypto extends DecryptResult {
+export type DecryptResultPmcrypto = Omit<DecryptResult, 'signatures'> & {
+    signatures: (OpenPGPSignature | undefined)[];
```
twiss, line 133: When can one of the elements be undefined?

lib/message/decrypt.js
```diff
@@ -12,7 +12,7 @@ export async function decryptMessage(options) {
 
     try {
         const result = await openpgp.decrypt(options);
-        const { data, filename, verified, signatures, errors } = await handleVerificationResult(
+        const { data, filename, verified, errors, signatures } = await handleVerificationResult(
```
twiss, line 15: This becomes very nitpicky but... why? This looks stranger than what it was IMO. Also below (`errors, signatures: pgpSignatures`).

econdepe: true, it's Frankestein code

lib/index.d.ts
econdepe: Never, my bad

PR #81, opened 2020-04-29
always return an array of signatures when decrypting a message

econdepe:
#80 was incomplete

### DIFF
--- a/lib/message/processMIME.js
+++ b/lib/message/processMIME.js
@@ -29,7 +29,7 @@ const verifySignature = async ({ publicKeys, date }, data) => {
     }
     const [, rawboundary] = /boundary\s*=\s*([^;]*)\s*(;|$)/gi.exec(contentType) || [];
     if (!rawboundary) {
-        return data;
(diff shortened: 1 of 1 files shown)


lib/message/processMIME.js
```diff
@@ -154,5 +154,5 @@ const parse = async (
 export default async function processMIME(options, data) {
     const { subdata, verified, signatures } = await verifySignature(options, data);
 
-    return parse(options, subdata, verified, signatures);
+    return parse(options, subdata, verified, signatures || []);
```
twiss, line 157: Shouldn't `verifySignature` always return a signatures array? Well.. except for
```js
    if (!rawboundary) {
        return data;
    }
```
but I think that should be 
```js
        return { subdata: data, verified: 0, signatures: [] };
```
Other than that, LGTM :+1:

econdepe: That sounds like the logic thing, but I ignored if
```
if (!rawboundary) {
    return data;
}
```
would have a deeper meaning :man_shrugging: . But it looks more like a typo, so let's do that!

PR #93, opened 2020-07-23
 Fix #92 E-mail ending matching for keyCheck()

### LINKED ISSUE #92: Escape RegExp characters in e-mail in keyCheck() function
Hello!

By looking at the code at https://github.com/ProtonMail/pmcrypto/blob/master/lib/key/check.js#L18 I assume that if `info.user.userId` ends with `"<jackyblack@foo.com>"` and `email` is `"jack.black@foo.com"`, it would match as `.` is any character in regular expression.

I think this would trigger unwanted behaviors.

If my analyse sounds correct to you, you can assign this issue to me and I will propose a PR to escape this `email` variable.

kylekatarnls:
Previously: `keyCheck(info, 'jack.black@foo.com')` threw no error with `userId: 'Jacky Black <jackyblack@foo.com>'`
With this change it throw `UserID does not contain correct email address`

Fix #92

### DIFF
--- a/lib/key/check.js
+++ b/lib/key/check.js
@@ -15,7 +15,7 @@ export function keyCheck(info, email, expectEncrypted) {
             throw new Error('Missing or too many UserID packets');
         }
 
-        if ((info.user.userId + '').substr(-2 - email.length) !== '<' + email + '>') {
+        if ((info.user.userId + '').endsWith('<' + email + '>')) {
             throw new Error('UserID does not contain correct email address');
         }
     }
--- a/lib/key/check.js
+++ b/lib/key/check.js
@@ -15,7 +15,7 @@ export function keyCheck(info, email, expectEncrypted) {
             throw new Error('Missing or too many UserID packets');
         }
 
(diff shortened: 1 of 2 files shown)


twiss:
Hey :wave: @dhoko corrected me, apparently we're automatically adding polyfills in pmcrypto, unlike in OpenPGP.js. So feel free to revert to your first commit. Sorry for the back and forth!

lib/key/check.js
```diff
@@ -15,7 +15,7 @@ export function keyCheck(info, email, expectEncrypted) {
             throw new Error('Missing or too many UserID packets');
         }
 
-        if (!new RegExp('<' + email + '>$').test(info.user.userId)) {
+        if ((info.user.userId + '').endsWith('<' + email + '>')) {
```
twiss, line 18: Thanks :blush: A few tests failed, one of them is because of this:
```suggestion
        if (!(info.user.userId + '').endsWith('<' + email + '>')) {
```
The other test failures are unrelated and should be gone after merging #91.

twiss:
@kylekatarnls Thanks for the patience, and thanks again for the PR!

PR #97, opened 2020-09-23
Added support for encrypted detached signatures

Cronosh:
When encrypting a message with detached signature, now the detached encrypted signature is available to use. If a detached encrypted signature exists, decryption decrypts and uses that one.

### DIFF
--- a/lib/message/encrypt.js
+++ b/lib/message/encrypt.js
@@ -3,7 +3,7 @@ import { openpgp } from '../openpgp';
 import { serverTime } from '../serverTime';
 import { createMessage } from './utils';
 
-export default function encryptMessage(options) {
+export default async function encryptMessage(options) {
     if (typeof options.data === 'string') {
         options.message = createMessage(openpgp.util.removeTrailingSpaces(options.data), options.filename);
     }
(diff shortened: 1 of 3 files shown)


lib/message/encrypt.js
```diff
@@ -15,5 +15,24 @@ export default function encryptMessage(options) {
...
+        // Create detached signature of message
+        const signature = await options.message.signDetached(options.privateKeys);
+
+        // Encrypt message without signing it (thus no need to change .detached)
+        options.privateKeys = [];
+        const result = await openpgp.encrypt(options);
+
+        // Encrypt signature and add it to the final result
+        options.message = createMessage(signature.packets.write(), options.filename);
+        const encryptedSignature = await openpgp.encrypt(options);
```
twiss, line 28: I think `options.filename` isn't needed here, as it refers to the filename of the data, not of the signature.
Also, perhaps instead of overwriting the original `options` object, you could create a new one here? Something like
```suggestion
        const encryptedSignature = await openpgp.encrypt({
            ...options,
            message: createMessage(signature.packets.write())
        });
```

lib/message/encrypt.js
```diff
@@ -15,5 +15,24 @@ export default function encryptMessage(options) {
...
     options.compression = options.compression ? openpgp.enums.compression.zlib : undefined;
 
-    return openpgp.encrypt(options);
+    if (options.detached) {
+        // Create detached signature of message
+        const signature = await options.message.signDetached(options.privateKeys);
+
+        // Encrypt message without signing it (thus no need to change .detached)
+        options.privateKeys = [];
+        const result = await openpgp.encrypt(options);
```
twiss, line 24: And then also here, you could do
```suggestion
        const result = await openpgp.encrypt({
            ...options,
            privateKeys: []
        });
```

lib/message/encrypt.js
```diff
@@ -15,5 +15,24 @@ export default function encryptMessage(options) {
...
+
+        // Encrypt signature and add it to the final result
+        options.message = createMessage(signature.packets.write(), options.filename);
+        const encryptedSignature = await openpgp.encrypt(options);
+        result.encryptedSignature = options.armor ? encryptedSignature.data : encryptedSignature.message;
```
twiss, line 29: Oh, and finally, this doesn't quite work (your test currently fails) because `options.armor` is undefined by default in this function, whereas it's true by default inside `openpgp.encrypt`. So you have to change this function's signature to something like `function encryptMessage({ armor = true, ...options })`, and then pass `armor` back into the calls to `openpgp.encrypt`.

Cronosh:
Ok thanks a lot for the comments. The last commit addresses them.

PR #99, opened 2020-09-29
Added support for detached signatures encrypted with session keys and tests on returnSessionKey

Cronosh:


lib/message/encrypt.js
```diff
@@ -15,30 +15,26 @@ export default async function encryptMessage({ armor = true, ...options }) {
...
         // Encrypt signature and add it to the final result
-        const encryptedSignature = await openpgp.encrypt({
-            ...options,
-            message: createMessage(signature.packets.write()),
-            privateKeys: [],
-            armor
-        });
-        result.encryptedSignature = armor ? encryptedSignature.data : encryptedSignature.message;
+        options.message = createMessage(signature.packets.write());
+        options.sessionKey = result.sessionKey;
```
twiss, line 30: This works if `options.returnSessionKey` is set, but breaks things if `options.sessionKey` is set but `options.returnSessionKey` isn't, I think. You could force `returnSessionKey = true` before encrypting the message above, if you want - I think it shouldn't do any harm. If you want to be super safe about things, you could add another test where you pass `sessionKey` but not `returnSessionKey`.

lib/message/encrypt.js
```diff
@@ -16,6 +16,7 @@ export default async function encryptMessage({ armor = true, ...options }) {
     options.compression = options.compression ? openpgp.enums.compression.zlib : undefined;
 
     options.armor = armor;
+    options.returnSessionKey = returnSessionKey;
```
twiss, line 19: Instead of making it a default and setting it here, I would just set `options.returnSessionKey = true`, but below instead of here (inside the `if (options.detached) {`, since that's when you need it).

test/message/encryptMessage.spec.js
```diff
@@ -102,3 +102,113 @@ test('it can encrypt a message and decrypt it unarmored using session keys along
     t.is(decrypted, 'Hello world!');
     t.is(verified, VERIFICATION_STATUS.SIGNED_AND_VALID);
 });
+
+test('it can encrypt and decrypt a message with session key without setting returnSessionKey', async (t) => {
+    const decryptedPrivateKey = await decryptPrivateKey(testPrivateKeyLegacy, '123');
+    const sessionKey = {
+        data: new Uint8Array([
+            197,
```
twiss, line 110: I know that the linter wants it like this, but I think allocating 32 lines of code for a session key is pretty ugly :) Perhaps you could convert it to hex (in the code) and then do `data: openpgp.util.hex_to_Uint8Array('...')`

PR #100, opened 2020-10-09
[WIP] Introducing OpenPGPjs v5

Cronosh:


lib/utils.js
```diff
@@ -16,46 +26,154 @@ export const decodeBase64 = ifDefined((input) => localAtob(input.trim()));
...
+        if (isStream(data)) {
+            const hashInstance = new Hash();
+            return transform(
+                data,
+                (value) => {
+                    hashInstance.process(value);
+                },
+                () => hashInstance.finish().result
+            );
+        }
```
twiss, line 138: I think you don't need to support streams here

lib/utils.js
```diff
@@ -16,46 +26,154 @@ export const decodeBase64 = ifDefined((input) => localAtob(input.trim()));
...
+        return Hash.bytes(data);
+    };
+}
+
+export const SHA256 = (args) => asmcryptoHash(Sha256)(args);
+export const SHA512 = (args) => asmcryptoHash(Sha512)(args);
```
twiss, line 144: ... so you could replace this with just
```js
export const SHA256 = Sha256.bytes;
```
etc.

test/key/utils.spec.js
```diff
@@ -9,9 +9,13 @@ import {
...
     stripArmor
 } from '../../lib/pmcrypto';
-import { openpgp } from '../../lib/openpgp';
+import { enums } from 'openpgp';
 import { keyCheck } from '../../lib';
 
+const { Crypto } = require('@peculiar/webcrypto');
+
+global.crypto = new Crypto();
+
```
twiss, line 18: This should be redundant now :)

test/helper.js
```diff
@@ -1,3 +1,4 @@
+import './_cryptoPolyfill';
```
twiss, line 1: Out of curiosity, what does ava complain about without the underscore?

Cronosh: It was complaining that cryptoPolyfill didn't contain any test. According to [https://github.com/avajs/ava/blob/master/docs/05-command-line.md](ava's documentation) there are a bunch of ways to tell ava to ignore a file. "helpers" are ignored by default and so are file starting with _. Another way might be to specify the file in package.json but that sounded a bit overkilling to me.

twiss: I see, thanks :+1:

larabr:
Superseeded by #123

PR #103, opened 2020-10-22
Add promise on the return type of decryptMIMEMessage

Swiip:
decryptMIMEMessage is async and returns a promis, this PR should align types to this.

Yet, I have hard time to understand the api design of decryptMIMEMessage.

It returns an object of async function, it would have sense if the function was indeed synchronous. But it's not, the function is async, wait for everything to be ready anyway.

Perhaps an historic choice? But feels strange today.

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -149,14 +149,14 @@ export function decryptMessageLegacy(options: DecryptLegacyOptions): Promise<Dec
 
 export function decryptMIMEMessage(
     options: DecryptMimeOptions
-): {
+): Promise<{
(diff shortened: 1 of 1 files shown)


twiss:
Yeah, this is historic. There used to be a reason for this (although I forget exactly what :stuck_out_tongue:) but now it's just for backwards compatibility.

That being said, there could be again a use-case for this in the future, if we consider streaming: even though this function doesn't support streaming today, in the future, we could e.g. make it return
```js
Promise<{
  getBody: () => ReadableStream<string>
  verify: () => Promise<number>
}>
```
which is close to what OpenPGP.js returns in the streaming case today (since the data is available before the verification result).

PR #110, opened 2021-01-15
Set creation time of generated keys 1 minute in the past

Cronosh:
Add an `offset` parameter to `generateKey`, which is a number of milliseconds which gets added to the current (server) time, defaulting to minus one minute, to help avoid race conditions from clock desync.

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -10,3 +10,5 @@ export const SIGNATURE_TYPES = {
 };
 
 export const MAX_ENC_HEADER_LENGTH = 1024;
+
+export const DEFAULT_OFFSET = -86400000;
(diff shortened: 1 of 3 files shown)


lib/constants.js
```diff
@@ -10,3 +10,5 @@ export const SIGNATURE_TYPES = {
 };
 
 export const MAX_ENC_HEADER_LENGTH = 1024;
+
+export const DEFAULT_OFFSET = -86400000;
```
twiss, line 14: I know I said this and Android does the same thing, but maybe if we're going to do this by default, we should default to a more reasonable offset, like 1 minute or 10 minutes or so?

Cronosh: Sure, no problem. Since the original problem is in the order of magnitude of seconds, 10 minutes is probably safer than 1?

twiss: Sure :+1: Even though I feel like if a server is off by a full minute, something has gone pretty wrong (they should all have ntp I think)

Cronosh: 1 minute it is then!

twiss (approved):

PR #111, opened 2021-04-20
Fix `verifyMessage` return type definitions

larabr:
`verifyMessage` does not return the same `signatures` object as `openpgpjs.verify`, since `handleVerificationResult` extracts the `signature` field from each entry:

- https://github.com/ProtonMail/pmcrypto/blob/55021f777aa775c581feda90beae302dc0272b27/lib/message/utils.js#L164
- https://github.com/ProtonMail/pmcrypto/blob/55021f777aa775c581feda90beae302dc0272b27/lib/message/utils.js#L101


twiss:
Superseded by #113

PR #115, opened 2021-04-27
Throw on server time updates too far off from local time

larabr:
Throw error on server time updates that are not within 24 hours of the local time.

- [ ] should we throw on `null` serverTime, rather than defaulting to local time?


larabr:
This check will be implemented directly in the webapp.

PR #117, opened 2021-07-01
Don't consider keys that cannot encrypt as expired, add `isRevokedKey()` and `canKeyEncrypt()`

larabr:
- Change `isExpiredKey()` behaviour:
  - consider primary key expiration instead of encryption key expiration (i.e. now passing a verify-only key won't always return `false`)
  - do not check for key revocation
- Change `keyInfo()` and `createPacketInfo()` behaviour: the `expires` returned field now contains the primary key expiration time, instead of the encryption key expiration

- Add `isRevokedKey()` to check for primary key revocation
- Add `canKeyEncrypt()` to check whether a key is valid for encryption
- Fix type definition for `OpenPGPKey.getExpirationTime()`


lib/index.d.ts
```diff
@@ -281,3 +285,5 @@ export function serverTime(): Date;
 export function getPreferredAlgorithm(key: OpenPGPKey[], date?: Date): Promise<string>
 
 export function getSHA256Fingerprints(key: OpenPGPKey): Promise<string[]>
+
+export function canKeyEncrypt(key: OpenPGPKey, date?: Date): Promise<boolean>;
```
twiss, line 289: Nit: newline :)

twiss (approved):

econdepe (approved):

PR #121, opened 2021-08-13
Add `checkKeyStrength`

larabr:


lib/key/check.js
```diff
@@ -1,6 +1,102 @@
...
+    const minRSABits = 1024;
+    const allowedCurves = new Set([
+        enums.curve.ed25519,
+        enums.curve.curve25519,
+        enums.curve.p256,
+        enums.curve.p384,
+        enums.curve.p521
+    ]);
+
+    // primary key only (TODO does this matter for contacts?)
```
larabr, line 79: ?

lib/key/check.js
```diff
@@ -1,6 +1,102 @@
...
+        allowedPublicKeyAlgorithms: new Set([
+            enums.publicKey.rsa_encrypt_sign,
+            enums.publicKey.rsa_sign,
+            enums.publicKey.ecdsa,
+            enums.publicKey.eddsa
+        ]),
+        allowedCurves: new Set([enums.curve.ed25519, enums.curve.p256, enums.curve.p384, enums.curve.p521]),
+        minRSABits: 2048
+    });
+    // TODO should we check all subkeys instead? (faster, but we dont know if they are signing or encryption subkeys)
```
larabr, line 49: ?

lib/index.d.ts
```diff
@@ -290,3 +290,5 @@ export function getPreferredAlgorithm(key: OpenPGPKey[], date?: Date): Promise<s
 export function getSHA256Fingerprints(key: OpenPGPKey): Promise<string[]>
 
 export function canKeyEncrypt(key: OpenPGPKey, date?: Date): Promise<boolean>;
+
+export function checkPersonalKeyStrength(key: OpenPGPKey): Promise<void>;
```
twiss, line 294: Newline please

lib/key/check.js
twiss: Yes, let's just check all subkeys, and say that all of them must be RSA/ECDSA/EdDSA/ECDH (without particularly looking at whether it's a signing or encryption key)

lib/key/check.js
twiss: Yes, for verifying signatures. IMHO, can't we just do the same thing here? IMHO, we should also warn for 1024-bit contact keys.

lib/key/check.js
```diff
@@ -1,6 +1,102 @@
...
+    // primary key
+    checkKeyStrength(publicKey, {
+        allowedPublicKeyAlgorithms: new Set([
+            enums.publicKey.rsa_encrypt_sign,
+            enums.publicKey.rsa_sign,
+            enums.publicKey.ecdsa,
+            enums.publicKey.eddsa
+        ]),
+        allowedCurves: new Set([enums.curve.ed25519, enums.curve.p256, enums.curve.p384, enums.curve.p521]),
+        minRSABits: 2048
```
twiss, line 47: Perhaps this should be 2047?

larabr: I'd do the change for OpenPGP.js v5 only?

lib/key/check.js
larabr: What same thing?
Contact key warnings are going to be shown in the composer, so I think we already agreed we should leave 2048 for now.

lib/key/check.js
twiss: Hmm, ok.. but then you're essentially relying on the bug that v4 returns 2047-bit keys as 2048, despite the intention being to accept 2047-bit keys, so I would just spell that out from the start

lib/key/check.js
twiss: OK, fair enough. I mean, I think eventually we should, but yeah we might need a transitional period where we don't since we generated those ourselves in the past. In any case, then I think the two functions can be essentially the same except for the `minRsaBits`, right?

lib/key/check.js
larabr: I never really considered it a bug, more of an intended rounding :D but yeah we can already put 2047 if you prefer.

lib/key/check.js
larabr: If we don't require encryption key presence, then yes, but since the use-case is different, I'd still expose two different functions? (so that the caller does not need to set params)

twiss (approved):

PR #123, opened 2021-08-25
update openpgpjs to version 5

Mandragorian:
TODO
- [ ] deprecate/rewrite keyInfo
- [ ] replace `isPrototype` with `instanceof`
- [ ] replace `typeof undefined`
- [ ] remove redundant `Promise.reject`
- [ ] check leftover (no longer thrown or redundant) error checks
- [ ] #126 ?
- [ ] Clean up TS defs to reuse those of OpenPGP.js (or rewrite in TS directly)
- [ ] Remove/strip down mailparser code
- [ ] use `PrivateKey` over `OpenPGPKey` where relevant
- [ ] use lightweight openpgp build (need to run tests in browser)

Non-breaking API changes:
- add `readKey()`
- add `generateSessionKeyFromKeyPreferences()`

Breaking changes:
- Require Webcrypto
- API changes
  - `init()`: no longer accepts an `openpgp` instance in input
  - `generateKey()`: remove `offset` param (the offset is accounted for in the default `date` option) 
  - `encryptMessage()` : 
     - removed `data`, `filename` input options
     - (temporary, to fix) no support for using detach-signature with streaming
  - `signMessage()`: removed `data` input option
  - `getPreferredAlgorithm()`: removed (to generate a session key based on key preferences, use `generateSessionKeyFromKeyPreferences()` directly)
  - `cloneKey()`, `getMaxConcurrency()`, `config`: removed (not needed)
  - `genPublicEphemeralKey()`, `genPrivateEphemeralKey()`: removed (unused in apps)
  - `createWorker()`: removed (to be replaced)

Superseeds #100


lib/message/encrypt.js
```diff
@@ -3,7 +3,11 @@ import { openpgp } from '../openpgp';
 import { serverTime } from '../serverTime';
 import { createMessage } from './utils';
 
-export default async function encryptMessage({ format = 'armored', ...options }) {
+export default async function encryptMessage({ armor = true, detached = false, ...options }) {
+    if (typeof options.format === 'undefined') {
+        options.format = armor ? 'armored' : 'object';
+    }
+
     if (typeof options.data === 'string') {
```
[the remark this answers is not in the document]
larabr: Does it make sense to leave the `armor` flag at all? I'd be for requiring the same `format` option as OpenPGP.js.

lib/utils.js
```diff
@@ -16,36 +18,146 @@ export const decodeBase64 = ifDefined((input) => localAtob(input.trim()));
...
+const getWebCrypto = () => {
+    return typeof globalThis !== 'undefined' && globalThis.crypto;
+};
+
+export const SHA256 = async (args) => {
+    const { subtle } = getWebCrypto();
+    const res = await subtle.digest('SHA-256', args);
+    const ar = new Uint8Array(res);
+    return ar;
+};
```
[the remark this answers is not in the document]
larabr: The tests are run in node though, so we should keep a compatible interface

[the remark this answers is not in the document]
twiss: In the tests we also use Web Crypto now, see https://github.com/ProtonMail/pmcrypto/pull/123/files#diff-a1e8c0de01ccdb3e6fe696c0c315f29c0bb8f978fc71647dec2869ff841cc06dR6

lib/key/utils.js
```diff
@@ -79,15 +121,15 @@ export async function isRevokedKey(key, date = serverTime()) {
...
     } catch (e) {
         return false;
     }
 };
 
 export async function compressKey(armoredKey) {
-    const [k] = await getKeys(armoredKey);
+    const k = await getKey(armoredKey);
     const { users } = k;
     users.forEach(({ otherCertifications }) => (otherCertifications.length = 0));
```
larabr, line 134:
```suggestion
    for (const user of users) { user.otherCertifications = []; };
```

lib/key/utils.js
```diff
@@ -97,29 +139,45 @@ export function getFingerprint(key) {
...
+function keyIDCheck(keyID, key) {
+    if (keyID.length === 16) {
+        return keyID === key.getKeyID().toHex();
+    }
+    return keyID === key.getFingerprint();
+}
+
+function getKeyWithID(keyID, keys) {
+    for (let i = 0; i < keys.length; i++) {
+        if (keyIDCheck(keyID, keys[i])) {
```
larabr, line 151: Can probably just use `key.getKeys(keyID)`

lib/key/utils.js
```diff
@@ -97,29 +139,45 @@ export function getFingerprint(key) {
...
-        loadPublic: () => keys,
-        loadPrivate: () => [],
-        storePublic() {},
-        storePrivate() {}
-    });
-
-    await keyring.load();
-
-    const keyids = signature.packets.map(({ issuerKeyId }) => issuerKeyId.toHex());
+    const keyids = signature.packets.map(({ issuerKeyID }) => issuerKeyID.toHex());
```
larabr, line 174: `signature.getSigningKeyIDs`?

lib/openpgp.js
```diff
@@ -1,17 +1,11 @@
...
-export { instance as openpgp };
-
-export const setInstance = (value) => {
-    instance = value;
-};
-
-export const setConfig = (openpgp) => {
-    openpgp.config.s2k_iteration_count_byte = 96;
+export const setConfig = () => {
+    config.s2k_iteration_count_byte = 96;
```
larabr, line 4: Use new config + write meaningful tests

test/key/info.spec.js
```diff
@@ -123,19 +125,23 @@ test('valid key', async (t) => {
 });
 
 test('newly generated RSA key', async (t) => {
-    const { publicKeyArmored } = await pmcrypto.generateKey({ userIds: [{}], passphrase: 'test' });
-    const { validationError } = await pmcrypto.keyInfo(publicKeyArmored);
+    config.preferredCompressionAlgorithm = enums.compression.zlib;
```
larabr, line 128: Reset config

test/message/util.spec.js
```diff
@@ -46,55 +46,52 @@ hhwl92GFuTXz4AEAqn4L+ayYgphejF/ZTRIseHPK+t521CT6NZKoVaHnTWQA
...
-    signaturePackets.forEach(({ verified }) => {
-        t.is(verified, true);
-    });
     t.is(signatureTimestamp, signaturePackets[0].created);
 });
 
 test('it verifies a message with multiple signatures and returns the timestamp of the valid signature', async (t) => {
-    const publicKey1 = (await openpgp.key.readArmored(armoredPublicKey)).keys[0];
-    const publicKey2 = (await openpgp.key.readArmored(armoredPublicKey2)).keys[0];
+    const publicKey1 = await readKey({ armoredKey: armoredPublicKey });
```
larabr, line 65: `getKey`?

lib/key/utils.js
```diff
@@ -1,51 +1,93 @@
...
     }
     const offsetDate = new Date(date.getTime() + offset);
-    return openpgp.generateKey({ passphrase, date: offsetDate, ...rest });
+    return openpgpGenerateKey({ passphrase, date: offsetDate, ...rest });
 }
 
-export function generateSessionKey(algorithm) {
-    return openpgp.crypto.generateSessionKey(algorithm);
+export function generateSessionKey(encryptionKey) {
+    return openpgpGenerateSessionKey({ encryptionKeys: [encryptionKey] });
```
larabr, line 67: Revert

lib/index.d.ts
```diff
@@ -26,16 +28,9 @@ export enum SIGNATURE_TYPES {
     CANONICAL_TEXT = 1
 }
 
-// type defined in OpenPGP is not complete
-export interface OpenPGPKey extends key.Key {
-    users?: { userId?: { userid?: string } }[];
-    getExpirationTime( // override upstream definition, since it does not properly declare `Infinity` as return type
-        capabilities?: "encrypt" | "sign" | "encrypt_sign", keyId?: type.keyid.Keyid, userId?: object
```
larabr, line 33: verify `capabilities` is correctly replaced here and in the apps (`getExpirationTime` has changed behaviour)

lib/key/utils.js
```diff
@@ -130,8 +188,7 @@ export async function getMatchingKey(signature, keys) {
 }
 
 export async function cloneKey(inputKey) {
-    const [key] = await getKeys(inputKey.toPacketlist().write());
-    return key;
+    return getKey(inputKey.toPacketList().write());
```
larabr, line 191: This is not needed anymore (?), thanks to the cloning performed by `openpgp.encryptKey`

lib/key/encrypt.js
```diff
@@ -8,20 +8,18 @@ export function encryptPrivateKey(inputKey, privKeyPassCode) {
             return Promise.reject(new Error('Missing private key passcode'));
```
larabr, line 8: No need for cloning

lib/index.d.ts
```diff
@@ -184,24 +174,24 @@ type EncryptOptionsPmcrypto = EncryptOptionsPmcryptoWithData | EncryptOptions;
...
-): Promise<EncryptResult<string, undefined, string, string>>;
+): Promise<EncryptResult<undefined, string, Uint8Array, string>>;
 export function encryptMessage(
     options: EncryptOptionsPmcrypto & { armor: false; detached?: false }
-): Promise<EncryptResult<undefined, message.Message>>;
+): Promise<EncryptResult<undefined, OpenPGPMessage>>;
 export function encryptMessage(
     options: EncryptOptionsPmcrypto & { armor: false; detached: true }
-): Promise<EncryptResult<undefined, message.Message, OpenPGPSignature, message.Message>>;
+): Promise<EncryptResult<undefined, OpenPGPMessage, Uint8Array, OpenPGPMessage>>;
```
larabr, line 186: Do we want the change in output types?

lib/key/utils.js
```diff
@@ -1,51 +1,93 @@
...
 
-export function generateSessionKey(algorithm) {
-    return openpgp.crypto.generateSessionKey(algorithm);
+export function generateSessionKey(encryptionKey) {
+    return openpgpGenerateSessionKey({ encryptionKeys: [encryptionKey] });
 }
 
 export async function getPreferredAlgorithm(keys, date = serverTime()) {
-    return openpgp.enums.read(openpgp.enums.symmetric, await openpgp.key.getPreferredAlgo('symmetric', keys, date));
+    return enums.read(enums.symmetric, await getPreferredAlgo('symmetric', keys, date));
```
larabr, line 71: I think we can remove this (and `getPreferredAlgo`) and have the apps call `generateSessionKeys` directly (no other usage so far)

lib/index.d.ts
```diff
@@ -82,7 +77,7 @@ export interface BinaryResult {
     data: Uint8Array;
     filename?: string;
     signatures?: {
-        keyid: type.keyid.Keyid;
+        keyid: KeyID;
         verified: Promise<boolean>;
         valid: boolean;
```
larabr, line 82: Is `valid` still there? (removed in v5)
Probably best to reference the openpgp.js SignResult type directly

lib/index.d.ts
```diff
@@ -137,20 +127,20 @@ export function encryptSessionKey(options: {
     passwords?: any[];
```
larabr, line 127: Reuse openpgp def

lib/key/decrypt.js
```diff
@@ -1,5 +1,5 @@
-import { openpgp } from '../openpgp';
-import { getKeys } from './utils';
+import { decryptKey, decryptSessionKeys } from 'openpgp';
+import { getKey } from './utils';
 
 export async function decryptPrivateKey(privKey, privKeyPassCode) {
```
larabr, line 4: Rename params

lib/utils.js
```diff
@@ -16,48 +18,145 @@ export const decodeBase64 = ifDefined((input) => localAtob(input.trim()));
...
+
+export const arrayToBinaryString = (bytes) => {
+    bytes = new Uint8Array(bytes);
+    const result = [];
+    const bs = 1 << 14;
+    const j = bytes.length;
+
+    for (let i = 0; i < j; i += bs) {
+        result.push(String.fromCharCode(...bytes.subarray(i, i + bs < j ? i + bs : j)));
```
larabr, line 66:
```suggestion
        result.push(String.fromCharCode.apply(String, bytes.subarray(i, i + bs < j ? i + bs : j)));
```

package.json
```diff
@@ -45,9 +49,12 @@
   },
   "homepage": "https://github.com/ProtonMail/pmcrypto#readme",
   "dependencies": {
+    "@openpgp/asmcrypto.js": "^2.3.2",
+    "@openpgp/tweetnacl": "^1.0.3",
+    "@openpgp/web-stream-tools": "0.0.6",
     "@types/openpgp": "^4.4.12",
     "esm": "^3.0.84",
-    "openpgp": "^4.10.8"
+    "openpgp": "5.0.0"
```
larabr, line 57: `~5.0.0` ?

lib/message/decrypt.js
```diff
@@ -1,34 +1,39 @@
...
+
     try {
         // If encryptedSignature exists, decrypt and use it
-        if (options.encryptedSignature) {
-            const decryptedSignature = await openpgp.decrypt({
+        if (encryptedSignature) {
+            const decryptedSignature = await decrypt({
                 ...options,
-                message: options.encryptedSignature,
+                message: await getMessage(encryptedSignature),
```
larabr, line 23: `encryptedSignature` should already be given parsed (same assumption as the input message)

lib/message/decrypt.js
```diff
@@ -84,12 +89,7 @@ export async function decryptMessageLegacy(options) {
     const params = { verified: VERIFICATION_STATUS.NOT_SIGNED, signatures: [], errors };
 
     // OpenPGP CFB mode with resync (https://tools.ietf.org/html/rfc4880#section-13.9)
-    const result = await openpgp.crypto.cfb.decrypt(
-        'aes256',
-        randomKey,
-        oldEncMessage.subarray(openpgp.crypto.cipher.aes256.blockSize + 2),
-        oldEncMessage.subarray(2, openpgp.crypto.cipher.aes256.blockSize + 2)
-    );
+    const result = await AES_CFB.decrypt(oldEncMessage.subarray(16 + 2), randomKey, oldEncMessage.subarray(2, 16 + 2));
```
larabr, line 92: Move this op to `crypto/` folder

lib/key/info.js
```diff
@@ -58,37 +59,37 @@ const primaryUser = async (key, date) => {
...
-    const mainFingerprint = keys[0].getFingerprint();
+    const algoInfo = key.getAlgorithmInfo();
+    const mainFingerprint = key.getFingerprint();
 
     const obj = {
-        version: keys[0].primaryKey.version,
-        publicKeyArmored: keys[0].toPublic().armor(),
+        version: key.keyPacket.version,
+        publicKeyArmored: key.toPublic().armor(),
         fingerprint: mainFingerprint, // FIXME: deprecated, use fingerprints instead
```
larabr, line 78: remove `fingerprint`?

lib/index.d.ts
twiss: Maybe, if we anyway want to change from `armor` to `format` parameter, then it's reasonable to change the output type too, yeah?

larabr: Yes but not in this way

test/message/encryptMessage.spec.js
```diff
@@ -151,66 +160,59 @@ test('it can encrypt and decrypt a message with session key without setting retu
...
     const decryptedPrivateKey = await decryptPrivateKey(testPrivateKeyLegacy, '123');
     const { message: encrypted, sessionKey: sessionKeys, encryptedSignature } = await encryptMessage({
-        message: createMessage('Hello world!'),
-        publicKeys: [decryptedPrivateKey.toPublic()],
-        privateKeys: [decryptedPrivateKey],
-        streaming: 'web',
-        armor: false,
+        message: await createMessage('Hello world!'),
+        encryptionKeys: [decryptedPrivateKey.toPublic()],
+        signingKeys: [decryptedPrivateKey],
```
larabr, line 186: Re-enable streaming

lib/message/processMIME.js
```diff
@@ -49,8 +49,8 @@ const verifySignature = async ({ publicKeys = [], date }, data) => {
     const body = parts[1];
 
     const { data: subdata, verified, signatures } = await verifyMessage({
-        message: createCleartextMessage(body),
-        publicKeys,
+        message: await createMessage(body),
```
larabr, line 52: Is this equivalent?

twiss: No, it should probably stay as it was :sweat_smile:

lib/message/decrypt.js
```diff
@@ -1,34 +1,39 @@
...
 
         let verified = verificationResult.then((result) => result.verified);
         let verifiedSignatures = verificationResult.then((result) => result.signatures);
         let errors = verificationResult.then((result) => result.errors);
 
-        if (!openpgp.stream.isStream(result.data)) {
+        if (!isStream(result.data)) {
             verified = await verified;
             verifiedSignatures = await verifiedSignatures;
             errors = await errors;
```
larabr, line 39: is `errors` still relevant?

test/message/util.spec.js
```diff
@@ -46,55 +46,52 @@ hhwl92GFuTXz4AEAqn4L+ayYgphejF/ZTRIseHPK+t521CT6NZKoVaHnTWQA
...
+    const publicKey2 = await readKey({ armoredKey: armoredPublicKey2 });
     const { data, verified, signatureTimestamp, signatures, errors } = await verifyMessage({
-        message: createMessage('hello world'),
-        signature: await openpgp.signature.readArmored(detachedSignatureFromTwoKeys),
-        publicKeys: [publicKey1, publicKey2]
+        message: await createMessage('hello world'),
+        signature: await readSignature({ armoredSignature: detachedSignatureFromTwoKeys }),
+        verificationKeys: [publicKey1, publicKey2]
     });
-    t.deepEqual(data, openpgp.util.str_to_Uint8Array('hello world'));
```
larabr, line 56: Mmm how has this changed?

lib/key/info.js
twiss: Yeah, I think we can :+1: In general I think this function should eventually be replaced with individual functions that return specific info, but ok, we can do that later.

package.json
twiss: Hmhm, or "^5.0.0" to be consistent with the others :shrug:

lib/index.d.ts
```diff
@@ -259,7 +259,7 @@ export interface VerifyMessageResult {
     errors?: Error[];
 }
 export interface VerifyMessageOptions extends VerifyOptions {
-    detached?: boolean;
+    detached?: boolean; // TODO what is this?
```
twiss, line 262: I think the web app passed this property to verify a detached signature, not realizing that it's unnecessary, and then someone added it here to make TS happy :sweat_smile:

lib/message/decrypt.js
twiss: No, can be removed

lib/index.d.ts
```diff
@@ -26,39 +31,12 @@ export enum SIGNATURE_TYPES {
...
-    userIds: UserID[];
-    passphrase: string;
-    keyExpirationTime?: number;
-    date?: Date;
-}
-export function reformatKey(
-    option: ReformatKeyOptions
-): Promise<{ key: key.Key; privateKeyArmored: string; publicKeyArmored: string; revocationCertificate: string }>;
+// TODO (?) these actually differ from 'openpgp' in that the passphrase is required
+export { generateKey, reformatKey }
```
larabr, line 39: Will fix by rewriting the functions in TS directly

lib/key/utils.js
larabr: We actually didn't use this fn anywhere, so I've removed it

lib/key/utils.js
larabr: Reverted and added a separate `generateSessionKeyFromKeyPreferences`

lib/message/decrypt.js
larabr: It might still make sense to keep it around, we'll remove in separate PR in case

lib/message/encrypt.js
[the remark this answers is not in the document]
larabr: separate PR for this

lib/message/encrypt.js
```diff
@@ -1,40 +1,76 @@
...
+            encryptionKeys: sanitizedOptions.encryptionKeys,
+            date: sanitizedOptions.date,
+            encryptionUserIDs: sanitizedOptions.encryptionUserIDs,
+            config: sanitizedOptions.config
+        });
+    }
 
-        // Encrypt message without signing it
-        options.privateKeys = [];
-        options.returnSessionKey = true;
```
larabr, line 25: Why would this default to true for detach-signing?

lib/message/decrypt.js
twiss: Sorry, I was also wrong, it's still needed I think as `handleVerificationResult` converts the `verified` promise to false and moves the error to this array

lib/message/encrypt.js
twiss: Because we use the same session key for encrypting the detached signature (`encryptedSignature`)

test/message/util.spec.ts
```diff
@@ -45,97 +44,102 @@ hhwl92GFuTXz4AEAqn4L+ayYgphejF/ZTRIseHPK+t521CT6NZKoVaHnTWQA
...
+    const publicKey2 = await readKey({ armoredKey: armoredPublicKey2 });
     const { data, verified, signatureTimestamp, signatures, errors } = await verifyMessage({
-        message: createMessage('hello world'),
-        signature: await openpgp.signature.readArmored(detachedSignatureFromTwoKeys),
-        publicKeys: [publicKey1, publicKey2]
+        message: await createMessage('hello world'),
+        signature: await readSignature({ armoredSignature: detachedSignatureFromTwoKeys }),
+        verificationKeys: [publicKey1, publicKey2]
     });
-    t.deepEqual(data, openpgp.util.str_to_Uint8Array('hello world'));
```
larabr, line 55: Should we always return `binary` data?

lib/message/decrypt.js
larabr: Yes, but we might change the result data anyway with the worker

test/message/util.spec.ts
twiss: No, I think we can change that, similarly to OpenPGP.js

lib/utils.js
```diff
@@ -16,48 +19,107 @@ export const decodeBase64 = ifDefined((input) => localAtob(input.trim()));
...
+ * Convert a native javascript string to a Uint8Array of utf8 bytes
+ * @param {String|ReadableStream} str - The string to convert
+ * @returns {Uint8Array|ReadableStream} A valid squence of utf8 bytes.
+ */
+export function stringToUtf8Array(str) {
+    const encoder = new TextEncoder('utf-8');
+    function process(value, lastChunk = false) {
+        return encoder.encode(value, { stream: !lastChunk });
+    }
+    return transform(str, process, () => process('', true));
```
twiss, line 111: I know this is what OpenPGP.js did, but it's not actually necessary to pass `stream` for encoding, as there are no partial strings (unlike partial UTF-8 code points for decoding). Also, the encoding parameter has been removed, it's always UTF-8 now:
```suggestion
    const encoder = new TextEncoder();
    return transform(str, value => encoder.encode(value));
```

lib/utils.js
```diff
@@ -16,48 +19,107 @@ export const decodeBase64 = ifDefined((input) => localAtob(input.trim()));
...
-export const createWorker = ({ path = '', n }) => {
-    if (!path) {
-        throw new Error('Path to worker required');
+/**
+ * Convert a Uint8Array of utf8 bytes to a native javascript string
+ * @param {Uint8Array|ReadableStream} utf8 - A valid squence of utf8 bytes
+ * @returns {String|ReadableStream} A native javascript string.
+ */
+export function utf8ArrayToString(utf8) {
+    const decoder = new TextDecoder('utf-8');
```
twiss, line 120: UTF-8 is the default now:
```suggestion
    const decoder = new TextDecoder();
```

package.json
```diff
@@ -49,22 +53,25 @@
   },
   "homepage": "https://github.com/ProtonMail/pmcrypto#readme",
   "dependencies": {
-    "@types/openpgp": "^4.4.18",
+    "@openpgp/asmcrypto.js": "^2.3.2",
+    "@openpgp/tweetnacl": "^1.0.3",
+    "@openpgp/web-stream-tools": "0.0.6",
     "esm": "^3.0.84",
-    "openpgp": "^4.10.8"
+    "openpgp": "~5.0.0"
```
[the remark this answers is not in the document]
larabr: Given that 5.1 will be a breaking change for instance, I'm not sure if it's safe?

lib/utils.js
```diff
@@ -16,48 +19,107 @@ export const decodeBase64 = ifDefined((input) => localAtob(input.trim()));
...
-export const stringToUtf8Array = (args) => openpgp.util.encode_utf8(args);
-export const utf8ArrayToString = (args) => openpgp.util.decode_utf8(args);
-export const binaryStringToArray = (args) => openpgp.util.str_to_Uint8Array(args);
-export const arrayToBinaryString = (args) => openpgp.util.Uint8Array_to_str(args); // eslint-disable-line new-cap
-export const arrayToHexString = (args) => openpgp.util.Uint8Array_to_hex(args); // eslint-disable-line new-cap
-export const concatArrays = (args) => openpgp.util.concatUint8Array(args);
+export const concatArrays = (args) => concatUint8Array(args);
+
+const isString = (data) => {
+    return typeof data === 'string' || String.prototype.isPrototypeOf.call(data);
```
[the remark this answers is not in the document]
larabr: I will clean up stuff like this in a different PR, I want to minimise changes in this one

lib/message/encrypt.js
```diff
@@ -1,40 +1,76 @@
...
+            signingKeys: sanitizedOptions.signingKeys,
+            signingKeyIDs: sanitizedOptions.signingKeyIDs,
+            signingUserIDs: sanitizedOptions.signingUserIDs,
+            date: sanitizedOptions.date,
+            config: sanitizedOptions.config,
+            format: 'object',
+            detached: true
+        });
+        result.signature = armor ? signature.armor() : signature;
+        if (message.fromStream) {
```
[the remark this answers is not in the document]
larabr: Because I didn't think of the problem when I originally wrote the code, and I wanted to keep this as future reference 😅

lib/pmcrypto.js
```diff
@@ -1,55 +1,44 @@
...
+import { setConfig } from './openpgp';
 import { encodeUtf8, decodeUtf8, decodeUtf8Base64, encodeUtf8Base64, encodeBase64, decodeBase64 } from './utils';
 
-export const init = (openpgp) => {
-    if (!openpgp) {
-        throw new Error('OpenPGP required');
-    }
-    setInstance(openpgp);
-    setConfig(openpgp);
+export const init = () => {
```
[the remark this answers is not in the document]
larabr: This is going to be quite confusing with IDE suggestion, as the `openpgp` argument will be displayed in the tooltip. Plus, as soon as we upgrade to TS, compilation will fail if they pass it. And since we'll expose the worker, I'm not sure if this will be used by the apps anymore.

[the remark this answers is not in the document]
twiss: We could do
```suggestion
export const init = function() {
    if (arguments.length) {
        throw new Error('Loading OpenPGP separately is no longer required');
    }
```

package.json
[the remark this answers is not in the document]
twiss: Well - 5.1 shouldn't break the high-level API in principle, so it should be fine (but in any case we'll probably also have to switch to ProtonMail/openpgpjs eventually so this should still change anyway)

lib/pmcrypto.js
```diff
@@ -1,55 +1,47 @@
...
-    if (!openpgp) {
-        throw new Error('OpenPGP required');
+export function init() {
+    if (arguments.length) {
+        throw new Error('Loading OpenPGP separately is no longer required');
     }
-    setInstance(openpgp);
-    setConfig(openpgp);
+    setConfig();
 };
```
twiss, line 9:
```suggestion
}
```

twiss (approved):

PR #124, opened 2021-09-28
Fix `createMessage` type declaration to accept binary data

larabr:


lib/index.d.ts
```diff
@@ -215,7 +215,7 @@ interface SignOptionsPmcryptoWithData extends Omit<SignOptions, 'message'> {
 type SignOptionsPmcrypto = SignOptionsPmcryptoWithData | SignOptions;
 
 export function createMessage(
-    text: string | ReadableStream<String> | message.Message,
+    data: string | ReadableStream<String> | Uint8Array | ReadableStream<Uint8Array>,
```
twiss, line 218: `ReadableStream<Uint8Array>` isn't actually properly supported; it will call `message.fromText` with it.

larabr: Should we fix the code to support binary streams?

twiss (approved):

Issue #112, opened 2021-04-21
Add tests for type definitions


larabr:
The exported types should be checked, similarly to what is done in OpenPGP.js (https://github.com/openpgpjs/openpgpjs/blob/master/test/typescript/definitions.ts).

Related to #111 .

PR #127, opened 2021-10-06
Fix type definitions of `decryptMessageLegacy` and `decryptMIMEMessage`

larabr:
Also, rewrite tests in typescript for better detection of type issues (related to #112).


lib/message/compat.js
```diff
@@ -4,6 +4,11 @@ const protonmailCryptoTailMessage = '---END ENCRYPTED MESSAGE---';
 const protonmailCryptoHeaderRandomKey = '---BEGIN ENCRYPTED RANDOM KEY---';
 const protonmailCryptoTailRandomKey = '---END ENCRYPTED RANDOM KEY---';
 
+/**
+ * Extract armored encrypted message from email
+ * @param {String|Object} EmailPM
```
twiss, line 9: I know the function kinda sort accepts it, but do we ever pass an object here? If not I don't really think we should start explicitly allowing it now. Maybe we can even get rid of the checks at some point
```suggestion
 * @param {String} EmailPM
```

lib/message/compat.js
```diff
@@ -14,6 +19,11 @@ export function getEncMessageFromEmailPM(EmailPM) {
     return '';
 }
 
+/**
+ * Extract (legacy, custom) armored encrypted random key from email
+ * @param {String|Object} EmailPM
```
twiss, line 24: Idem
```suggestion
 * @param {String} EmailPM
```

lib/message/decrypt.js
```diff
@@ -50,7 +50,7 @@ export async function decryptMessage(options) {
 }
 
 // Backwards-compatible decrypt message function
-// 'message' option must be a string!
+// `options.message` must be a string, to properly handle legacy messages (and avoid misusing this function)
```
twiss, line 53: As written, it kind of sounds like it only needs to be a string if you care about legacy messages, which is true - but I would rather change that and remove the checks, so that this function always throws for other types, so that we enforce this rather than write a comment about it.
```suggestion
// `options.message` must be a string
```

twiss (commented):
Some nitpicks below. Alternatively, I'm also fine with merging this PR as-is and then changing the below in a separate PR.

PR #133, opened 2022-01-27
[v7] Take `text/binaryData` as input of `encrypt/sign/verifyMessage`, remove `createMessage/CleartextMessage`

### LINKED ISSUE #126: Split up `createMessage` into `createBinaryMessage` and `createTextMessage`
Keeping a single `createMessage` function prevents us from supporting both streamed text and streamed binary data (see https://github.com/ProtonMail/pmcrypto/pull/124#discussion_r717609748).

larabr:
Changes: 
- Remove `createMessage` and `createCleartextMessage`
- Change `encryptMessage`, `signMessage`, `verifyMessage`:
  - accept input data as either `textData` or `binaryData` (instead of `data`).
  - a `message` input can no longer be passed.
- in `signMessage` and `encryptMessage`, `textData` is used as-is. Previously, if some string `data` was passed, `signMessage` would create a CleartextMessage automatically, whereas in `encryptMessage` trailing spaces were automatically removed (namely, the data was treated again as a CleartextMessage payload).

In `encryptMessage`, taking `text/binaryData` instead of a `message` is needed to support encrypt + detached sign with streamed input data.
In `signMessage` and `verifyMessage`, the change is motivated by how the functions are used in the web apps (namely, only to sign/verify detached data).

This also makes it easier to integrate the functions with a web worker, since the input options are all of transferable type.

Close #126.

<hr>

TODOs:

- [ ] Take `format` input option instead of `armor`, and only return `message` field (remove `data` in output) -> separate PR
- [x] Support streamed encrypt + detach-signed (if `format != 'object'`)
- [x] Rename `options.text, binary` to `options.textData/binaryData`


lib/index.d.ts
```diff
@@ -138,11 +136,21 @@ export function decryptMIMEMessage(
...
 }>;
 
-export interface EncryptOptionsPmcryptoWithData extends Omit<EncryptOptions, 'message'> {
-    data: Uint8Array | string;
-    returnSessionKey?: boolean;
+type MaybeStream<T extends Uint8Array | string> = T | ReadableStream<T>;
+export interface EncryptOptionsPmcryptoWithTextData extends Omit<EncryptOptions, 'message'> {
+    textData: MaybeStream<string>;
+    binaryData?: undefined;
+    stripTrailingSpaces?: boolean;
```
larabr, line 143: Are we leaving this here, or are we requiring the user to clean the input themselves if needed?
Same question for sign/verifyMessage.
Note this is not the same as a `asCleartextMessage` flag, since the created message is always of type `Message`.

lib/message/encrypt.js
```diff
@@ -30,40 +57,34 @@ export default async function encryptMessage({
...
-            // TODO clone input stream -- sanitizedOptions.message.fromStream ? await readMessage({ clone(<input data>) }) : sanitizedOptions.message;
-            throw new Error('Encrypting and detached-signing a message created from a stream is not supported yet');
+        if (!armor && isStream(textData || binaryData)) {
+            throw new Error('Unsupported detached signature when streaming data and requesting non-armored result');
         }
-        const signature = await sign({
-            message,
+
+        const detachedSignatureBinary = await sign({
+            message: await createMessageFromPassiveStreamClone({ [dataType]: data }),
```
twiss, line 65: I think the code becomes simpler if you inline the above functions, e.g.
```suggestion
            message: await createMessage({
                [dataType]: isStream(data) ? passiveClone(data) : data,
                ...
            }),
```
(Also, date and options are not passed here, but they are above, probably you wanted to add them here?)

lib/index.d.ts
larabr: Typescript doesn't allow setting both options though

lib/key/utils.js
```diff
@@ -99,7 +96,7 @@ export async function isRevokedKey(key, date = serverTime()) {
  */
 export const canKeyEncrypt = async (publicKey, date = serverTime()) => {
     try {
-        await encrypt({ message: await createMessage('test message'), encryptionKeys: publicKey, date });
+        await encrypt({ message: await createMessage({ text: 'test message' }), encryptionKeys: publicKey, date });
```
larabr, line 99: Could use `encryptMessage` instead

lib/message/encrypt.js
```diff
@@ -30,40 +57,34 @@ export default async function encryptMessage({
...
-                })
-            );
-        }
-        // Encrypt signature
+
         result.encryptedSignature = await encrypt({
             ...sanitizedOptions,
-            // TODO clone if streaming signature
-            message: await createMessage({ binary: signature.packets.write() })
+            message: await createMessage({ binary: cloneIfStream(detachedSignatureBinary) })
```
twiss, line 77: And then here
```suggestion
            message: await createMessage({
                binary: isStream(detachedSignatureBinary) ? clone(detachedSignatureBinary) : detachedSignatureBinary
            })
```
Also, `clone` is very cheap for non-stream binary data (it just returns a subarray, it doesn't actually clone the data, since the goal of `stream.clone` is to be able to read twice) so it's also fine to unconditionally call clone. But OK, it's also fine to leave this ternary

lib/message/encrypt.js
```diff
@@ -30,40 +57,34 @@ export default async function encryptMessage({
     }
 
     if (detached) {
-        // Add cleartext signature for backward compatibility
-        const { message } = sanitizedOptions;
-        if (message.fromStream) {
-            // TODO clone input stream -- sanitizedOptions.message.fromStream ? await readMessage({ clone(<input data>) }) : sanitizedOptions.message;
-            throw new Error('Encrypting and detached-signing a message created from a stream is not supported yet');
+        if (!armor && isStream(textData || binaryData)) {
+            throw new Error('Unsupported detached signature when streaming data and requesting non-armored result');
```
twiss, line 61:
```suggestion
        if (!armor && isStream(data)) {
            throw new Error('Creating unarmored detached signatures when streaming data is not supported');
```
Also, why is this not supported, actually? (Not that it's important, just curious)

lib/message/encrypt.js
```diff
@@ -30,40 +57,34 @@ export default async function encryptMessage({
     }
 
     if (detached) {
-        // Add cleartext signature for backward compatibility
-        const { message } = sanitizedOptions;
-        if (message.fromStream) {
-            // TODO clone input stream -- sanitizedOptions.message.fromStream ? await readMessage({ clone(<input data>) }) : sanitizedOptions.message;
-            throw new Error('Encrypting and detached-signing a message created from a stream is not supported yet');
+        if (!armor && isStream(textData || binaryData)) {
```
larabr, line 60:
```suggestion
        if (!armor && isStream(data)) {
```

lib/message/utils.js
```diff
@@ -63,26 +75,30 @@ export async function getCleartextMessage(message) {
 }
 
 /**
- * Create a cleartext message from a text.
- * @param {String|openpgp.cleartext.CleartextMessage} message
- * @return {openpgp.cleartext.CleartextMessage}
+ * Get a signed message from the given data.
+ * Exactly one of `textData`, `binaryData`, `cleartextMessageData` must be specified.
```
larabr, line 79:
```suggestion
 * Either `textData` or `binaryData` must be specified.
```

lib/index.d.ts
twiss: Right, ok. That's good at least :+1: I still think it would be more obvious to the apps why that is if they had to call the function themselves, but yeah, we can also have it as an option.

lib/message/encrypt.js
larabr: RIght now it's either armored or object output, and with the latter there was some issue with pulling the streams in the resulting signatures IIRC (we had a call about this and opted for not supporting this case for simplicity).

twiss: OK :+1:

lib/message/encrypt.js
larabr: I don't think we need to pass the `date` since encrypt/sign overwrite it? `options` was intended for `createMessage` options in case we kept exposing the function, but now it's not needed.

lib/index.d.ts
larabr: I'd keep it for now and remove it before release based on how much is needed in the apps. Also so that if we do remove it, we'll have a more visible commit explicitly removing support for trimming in the history.

twiss (approved):

lib/message/encrypt.js
twiss: Actually, no, encrypt/sign don't overwrite it. The date passed to encrypt/sign is used for the signature creation date, but the message creation date still comes from `createMessage`. So it's probably better to still pass it

lib/message/utils.js
```diff
@@ -63,26 +75,30 @@ export async function getCleartextMessage(message) {
...
     }
-    return openpgpCreateCleartextMessage({ text: message });
-}
-
-export function createMessage(source, filename, date = serverTime()) {
-    if (Uint8Array.prototype.isPrototypeOf(source)) {
-        return openpgp_createMessage({ binary: source, filename, date });
-    }
-    return openpgp_createMessage({ text: source, filename, date });
-}
```
twiss, line 82: https://github.com/ProtonMail/pmcrypto/pull/133#discussion_r793833253 also means that we should probably keep this wrapper function, actually, even if it's just for the `date = serverTime()` default :/

twiss (changes requested):

twiss (approved):

PR #136, opened 2022-02-07
[v7] Add support for binary output in `encrypt/signMessage`: replace `armor` option with `format`

larabr:
The change improves support for outputting serialised data, which is safer to use since it is independent of internal changes in OpenPGP.js objects, which can be tricky to use e.g. in some streaming cases.

Breaking changes:
- in `sign/encryptMessage`, the option `armor: boolean` is replaced by `format: 'armored' | 'binary' | 'object'`. `format: 'armored'` is still the default and it is equivalent to `armor: true`.
- in `encryptMessage`, `result.data` has been removed, and the function always returns encrypted data as `result.message`.

Type definition changes:
- fix types to correctly infer whether the output data is streamed or not.
- replace `ReadableStream` with `WebStream`, since the former is not compatible with OpenPGP.js type declarations, due to the `AsyncIterable` not being implemented by all browsers (yet).

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -138,6 +138,8 @@ export function decryptMIMEMessage(
 
 type MaybeStream<T extends Uint8Array | string> = T | WebStream<T>;
 type Data = string | Uint8Array;
+export { WebStream };
+
 export interface EncryptOptionsPmcryptoWithTextData<T extends MaybeStream<string>> extends Omit<EncryptOptions, 'message'> {
     textData: T;
     binaryData?: undefined;
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -12,7 +12,8 @@ import {
     generateKey,
     PrivateKey,
     SessionKey,
-    encryptSessionKey
+    encryptSessionKey,
+    WebStream
 } from 'openpgp/lightweight';
 
 export function init(): void;
@@ -45,8 +46,7 @@ export interface DecryptMimeOptions extends DecryptLegacyOptions {
 }
 
 // No reuse from OpenPGP's equivalent
-export interface EncryptResult<D = undefined, M = undefined, S = undefined, E = undefined> {
-    data: D;
+export interface EncryptResult<M = undefined, S = undefined, E = undefined> {
     message: M;
     signature: S;
     sessionKey: SessionKey;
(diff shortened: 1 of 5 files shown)


twiss:
> * in `encryptMessage`, `result.data` has been removed, and the function always returns encrypted data as `result.message`.

Probably we should move in the opposite direction, and always return `result.data`, for symmetry with `decryptMessage`?

larabr:
> > * in `encryptMessage`, `result.data` has been removed, and the function always returns encrypted data as `result.message`.
> 
> Probably we should move in the opposite direction, and always return `result.data`, for symmetry with `decryptMessage`?

But `decryptMessage` returns decrypted data, while `encryptMessage` returns an OpenPGP message, so I think it makes sense to keep separate names

twiss:
OK, but in OpenPGP.js we also had `.data` in the `encrypt` result for armored encrypted data, and `.message` for returned message objects. So we sort of had the convention that `.message` is a message object, and `.data` isn't. I think it would be a bit confusing to break that convention now, even though we don't really have that in OpenPGP.js anymore.

larabr:
But wouldn't returning `data` for an object break the convention anyway?
Also, OpenPGP.js now takes options such as `armoredMessage`, which in the worker will also be taken by e.g. `decryptMessage`. Instead, options such as `textData/binaryData` are used to specifically refer to non-OpenPGP stuff. So, to avoid confusion in what should be passed where, I really think it'd be clearer to separate the meaning of "message" and "data" at the logical level, rather than using it as a type indication.

twiss:
> But wouldn't returning `data` for an object break the convention anyway?

Yeah, I mean I would also be in favor of not supporting that at all, in pmcrypto.

> Also, OpenPGP.js now takes options such as `armoredMessage`, which in the worker will also be taken by e.g. `decryptMessage`. Instead, options such as `textData/binaryData` are used to specifically refer to non-OpenPGP stuff. So, to avoid confusion in what should be passed where, I really think it'd be clearer to separate the meaning of "message" and "data" at the logical level, rather than using it as a type indication.

But yeah, that's fair enough :+1:

twiss (approved):
(Just to document here, I'm fine with keeping `format: 'object'` in this PR, but I think we should remove it before releasing v7, if possible)

PR #138, opened 2022-02-16
[v7] Worker

larabr:
Implement Worker and corresponding Proxy interface that allows invoking pmcrypto functions without importing OpenPGP.js in the main thread. To achieve this:
  - OpenPGP Keys, Message, Signature objects are always passed and returned in serialized (armored or binary) form.
  - Keys are internally stored by the worker, which returns a transferable object to index the original key and includes some key metadata (e.g. creation time, expiration time, userIDs...).

TS changes:
- rely on type parameters instead of function overloading when declaring functions whose output types depend on input parameters (e.g. `signMessage`, `encryptMessage`). This is because using type parameters, it is possible to extend/reuse type definitions when wrapping a function, keeping narrow type inference. Instead, with overloading, only the first matching definition is applied at compile time, while the remaining overloaded definitions (even if narrower) are discarded and not considered further. 

Also:
- add TS support outside of tests;
- Do not export `utils` functions as part of `pmcrypto` (import `utils` module directly, to avoid pulling in openpgp as dependency)
- update ESLint and Prettify setup to support TS files. Some prettier default rules have changed in the new version, which results in some old code being reformatted.

TODO:
- [ ] add JSDoc comments to api and api.model so that they are shown by IDE when using CryptoWorker/Proxy.
- [ ] cleanup/refine API:
   - [x] do we need support for setting custom global config (in API constructor)? (no)
- [x] ~transpile & bundle index.ts (webpack)~ 
- [x] ~generate and expose types (TS + copy-files)~
we delegate TS compilation (if needed) to the apps and assume they support TS, since the pmcrypto utils functions also rely on polyfills which we'd need to export separately (as webpack chunks) together with the bundled index file.

<hr>

HOW-TO update from v6 to v7 worker:
- `splitMessage` not implemented, call `generateSessionKey` followed by `encryptSessionKey` to get symmetric/asymmetric keys (separately for each key and/or password). 
- `privateKey.toPublic()` no longer available. First export a serialized public key (`exportPublicKey(...)`) and then import the corresponding public key (`importPublicKey(...)`). Calling `importPublicKey` with a private key input is not recommended, since this edge case was implemented to allow importing encrypted private keys, so support for it might be dropped in the future.
- ...


lib/worker/transferHandlers.ts
```diff
@@ -0,0 +1,71 @@
...
+                serializedOptions[name].map(KeyReferenceSerializer.deserialize) :
+                KeyReferenceSerializer.deserialize(serializedOptions[name]);
+        }
+      });
+
+      return options;
+    }
+}
+
+export const customTransferHandlers = [{
```
larabr, line 51: Need to add `keyID` handler

lib/worker/worker.ts
```diff
@@ -0,0 +1,612 @@
...
+        }
+
+        return transfer(serialisedResult, buffers);
+    },
+    /**
+     * Generating a session key for the specified symmetric algorithm.
+     * To generate a session key based on some recipient's public key preferences,
+     * use `generateSessionKeyFromKeyPreferences()` instead.
+     */
+    generateSessionKey: async (algoName: Parameters<typeof generateSessionKey>[0]) => {
```
twiss, line 552: Can we swap these? I.e. `generateSessionKey` takes `{ recipientKeys }`, and call the other one `generateSessionKeyForAlgorithm`. As the latter is the more low-level one.

.eslintrc.json
```diff
@@ -68,6 +74,10 @@
         "eol-last": ["error"],
         "camelcase": ["error", {"allow": ["openpgp_*"]}],
 
+        "@typescript-eslint/naming-convention": ["error", {
+            "selector": "typeLike",
+            "format": ["PascalCase", "UPPER_CASE"]
+          }],
```
twiss, line 80:
```suggestion
        }],
```

lib/worker/worker.ts
```diff
@@ -0,0 +1,612 @@
...
+     */
+    importPrivateKey: async <T extends Data>({
+        armoredKey,
+        binaryKey,
+        passphrase
+    }: WorkerImportPrivateKeyOptions<T>) => {
+        const expectDecrypted = passphrase === null;
+        const maybeEncryptedKey = binaryKey ?
+            await readPrivateKey({ binaryKey }) :
+            await readPrivateKey({ armoredKey: armoredKey! });
```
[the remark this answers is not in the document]
larabr: OpenPGP.js (TS defs) does not allow setting both options

package.json
```diff
@@ -39,12 +40,13 @@
     },
     "homepage": "https://github.com/ProtonMail/pmcrypto#readme",
     "dependencies": {
+        "comlink": "^4.3.1",
+        "openpgp": "github:larabr/openpgpjs#ts-update-stream-types-and-other-options",
```
larabr, line 44: Declaring OpenPGP.js as non-dev dep is needed as long as we export utils functions which are in files that import `openpgp` modules. Otherwise webpack/TS in the apps fail to compile (before tree-shaking).

To avoid this, we should consider moving the util functions and constants to separate files or to the webapp repo directly.

lib/worker/worker.ts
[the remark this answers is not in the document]
twiss: Hmm. You're right, if it's an object literal - but if you just pass along an options object (i.e. don't destructure it first), it does work

package.json
```diff
@@ -2,20 +2,21 @@
...
     },
     "scripts": {
-        "test": "karma start karma.conf.js",
+        "build-worker": "webpack --config webpack.worker.js",
+        "build": "npm run build-worker",
+        "test": "npm run build-worker && karma start karma.conf.js",
         "lint": "eslint lib test --quiet",
-        "pretty": "prettier --write $(find lib -type f -name '*.js')",
+        "pretty": "prettier --write lib",
+        "preversion": "npm run build && git add -A dist",
```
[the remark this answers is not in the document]
larabr: As discussed, we are not going to distribute the pre-bundled worker in the end, since dealing with the large number of generated chunks would be complicated for the webapp.

lib/worker/api.ts
```diff
@@ -0,0 +1,750 @@
...
+    /**
+     * Each instance keeps a dedicated key storage.
+     * @param openpgpConfig - Config entries to set as OpenPGP global configuration.
+     *                        Note: functions whose behaviour depends on it accept a `config` input option, that can be used to
+     *                        apply a configuration change limited to the single function call.
+     * @throws {Error} if the given configuration is invalid
+     */
+    constructor(openpgpConfig: OpenPGPConfig = {}) {
+        super();
+        setStandardOpenPGPConfig();
```
larabr, line 302: Since the config is shared across the code in a worker/import (as they all use the same openpgp instance), in principle it makes more sense to set the custom config as part of `pmcrypto.init(openpgpCustomConfig)`. This would work for when the caller creates API instances directly (and not wrapped by a worker, like WorkerPool does).
However, `WorkerPool.init` then has no way to set the config for each worker (since it cannot access the corresponding pmcrypto instance).

lib/worker/transferHandlers/streamHandler.ts
```diff
@@ -10,16 +10,30 @@ export const ReadableStreamSerializer = {
     serialize: (readableStream: WebStream<Data>): MessagePort => {
         const { port1, port2 } = new MessageChannel();
 
-        const reader = readableStream.getReader();
+        // wait to get the reader until the first chunk is requested
```
larabr, line 13: Is this change needed?

lib/worker/api.ts
larabr: We are removing the option to set a custom global config, so this will be removed as well.

larabr:
Closing since the worker code has been moved to the web-clients repo.
All pmcrypto-related changes have been included in #141 ,

PR #139, opened 2022-03-15
[v7] `processMIME`: take `data` as part of `options`, uniform default values and add TS definitions

larabr:
Breaking changes to `processMIME`:
- drop second argument (`data`) and take `options.data` instead. This is to align the function interface with `encrypt/sign/verify/decryptMessage`.
- Rename `options.publicKeys` to `.verificationKeys` (follow up to #123).
- If no encrypted subject is found, `result.encryptedSubject` defaults to empty string instead of `false`. The app logic should be unaffected as `encryptedSubject` was already [declared](https://github.com/ProtonMail/WebClients/blob/d8cce0c430c3d51e3b611a53095ee6ce285594bc/applications/mail/src/app/helpers/message/messageDecrypt.ts#L38) and [treated](https://github.com/ProtonMail/WebClients/blob/35c0972864b13b54dd0be88dd84b5a1692f2d0e4/applications/mail/src/app/components/message/extras/ExtraDecryptedSubject.tsx) as a string in all cases.
- If the message does not have a body, the result now includes the fields `body: ''` and `mimetype: undefined` fields. Before, the fields were omitted. The change is again inline with the existing [webapp types](https://github.com/ProtonMail/WebClients/blob/d8cce0c430c3d51e3b611a53095ee6ce285594bc/applications/mail/src/app/helpers/message/messageDecrypt.ts#L57).

Also:
- `options.sender` defaults to empty string instead of `null` (otherwise the function would break in [this edge case](https://github.com/ProtonMail/pmcrypto/compare/v7...larabr:ts-processMIME?expand=1#diff-750da7b8a89bd39d114d6616655f302fa655511a4bc324d03a957decc6a1e5e2R118))
- default `options.date` to `serverTime()`
- add tests for MIME messages with attachment and encrypted subject;
- add TS definitions for `processMIME`

### DIFF
--- a/lib/message/processMIME.js
+++ b/lib/message/processMIME.js
@@ -161,11 +161,11 @@ const parse = async (
 /**
  * Process the mime structure in the right attachment.
  * @param {Object} options - options for signature verification and MIME processing
+ * @param {String} options.data - MIME data to process
  * @param {PublicKey[]} [options.verificationKeys]
  * @param {Date} [options.date] - to use for signature verification, instead of the server time
  * @param {String} [options.headerFilename] - the file name a memoryhole header should have
  * @param {String} [options.sender] - the address of the sender of this message
- * @param {String} data - MIME data to process
  * @return {Promise<{
  *      body: String | undefined,
  *      attachments: Object[],
@@ -175,7 +175,7 @@ const parse = async (
  *      signatures: Signature[]
  * }>}
  */
-export default async function processMIME(options, data) {
+export default async function processMIME({ data, ...options}) {
     const { subdata, verified, signatures } = await verifySignature(options, data);
 
     return parse(options, subdata, verified, signatures);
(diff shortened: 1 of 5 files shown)


lib/message/processMIME.js
```diff
@@ -64,21 +65,30 @@ const verifySignature = async ({ publicKeys = [], date }, data) => {
...
- * @param verified
- * @returns {Promise.<*>}
+ * @param {Object} options
+ * @param {String} [options.headerFilename] - The file name a memoryhole header should have
+ * @param {String} [options.sender] - the address of the sender of this message
+ * @param {String} [content] - mail content to parse
+ * @param {VERIFICATION_STATUS} [verified]
+ * @param {Signature[]} [signatures]
+ * @returns {Promise<{
+ *      body: String | undefined,
```
larabr, line 75:
```suggestion
 *      body: String,
```

PR #141, opened 2022-06-14
[v7] Various API changes and added support for TS modules

larabr:
See commit messages for list of changes


lib/utils.ts
```diff
@@ -0,0 +1,128 @@
+// Most of these util functions are copied as-is from https://github.com/openpgpjs/openpgpjs/blob/v5.0.0/src/util.js
+// @ts-ignore missing TS defs
+import { concatUint8Array, transform } from '@openpgp/web-stream-tools';
```
twiss, line 3: It would be best not to import `@openpgp/web-stream-tools` for this, if possible, as it also has a dependency on the stream polyfills. I think you could copy / adapt `concatUint8Array` from there; and then for `transform`, is the stream support of `stringToUtf8Array` and `utf8ArrayToString` used anywhere? If not, you can remove the `transform`s just do it in one step.

If it *is* used, it won't work like this, you need to call `loadStreamsPonyfill()` in v0.0.10 of web-stream-tools before using a `TransformStream`, as otherwise it won't be available in all browsers.

Another option would be to manually create a `ReadableStream` that when read, reads the original stream and processes each chunk manually. Lmk if you want help with that.

test/message/encryptMessage.spec.ts
```diff
@@ -74,11 +79,16 @@ describe('encryptMessage', () => {
 
     it('it does not compress a message by default', async () => {
         const decryptedPrivateKey = await decryptPrivateKey(testPrivateKeyLegacy, '123');
-        const { message: encrypted, sessionKey } = await encryptMessage({
+        const sessionKey = {
+            data: hexToUint8Array('c5629d840fd64ef55aea474f87dcdeef76bbc798a340ef67045315eb7924a36f'),
+            algorithm: enums.read(enums.symmetric, enums.symmetric.aes256)
+        };
```
twiss, line 85: This is fairly nitpicky but could we have a call to `generateSessionKey` here instead, to show how it "should" be done? (Idem in the other ones)

larabr: I can add an end-to-end test, but I prefer the encryptMessage tests to not depend on other functions for the most part 👍 
I'll add usage examples in the readme/other docs anyway.

lib/utils.ts
larabr: Even if it isn't used in the library, since pmcrypto supports streaming, I think we should have these utils support streams too... But as far as polyfilling goes, I'd let the client load them , it's what e.g. Drive already did.

Wouldn't using a ReadableStream still require an import?

test/message/encryptMessage.spec.ts
twiss: Alrighty :+1:

lib/utils.ts
larabr: Let's go for the ReadableStreams then, to avoid making the function `async`

twiss (approved):

PR #144, opened 2022-06-17
[v7] Fix `splitMessage` and clean up utils

larabr:
See commits for details.


lib/message/decrypt.js
```diff
@@ -49,30 +49,35 @@ export async function decryptMessage({ date = serverTime(), encryptedSignature,
 
 // Backwards-compatible decrypt message function
 // `options.message` must be a string, to properly handle legacy messages (and avoid misusing this function)
-export async function decryptMessageLegacy({ messageDate, ...options }) {
-    const sanitizedOptions = { ...options };
+export async function decryptMessageLegacy({ messageDate, message, decryptionKeys, ...options }) {
+    if (!isString(message)) {
+        throw new Error('Armored message expected. Use `decryptMessage` for binary messages or Message objects.')
+    }
```
twiss: Perhaps this should be called `armoredMessage`, then?

lib/message/decrypt.js
```diff
@@ -49,30 +49,35 @@ export async function decryptMessage({ date = serverTime(), encryptedSignature,
...
+    const oldEncRandomKey = getEncRandomKeyFromEmailPM(message);
 
-    // OpenPGP
+    // Standard OpenPGP message
     if (oldEncMessage === '' || oldEncRandomKey === '') {
         // Convert message string to object
-        sanitizedOptions.message = await getMessage(sanitizedOptions.message);
-        return decryptMessage(sanitizedOptions);
+        return decryptMessage({
+            message: await getMessage(message),
```
twiss: And then here, since we know it's an armored message, we can just do `readMessage({ armoredMessage })`, then we don't have to check the type again. In fact, can we get rid of `getMessage`, `getSignature` and `getCleartextMessage`? (Doesn't have to happen in this PR)

lib/utils.ts
```diff
@@ -92,6 +79,20 @@ export const arrayToBinaryString = (bytes: Uint8Array) => {
     return result.join('');
 };
 
+/**
+ * Convert a hex string to an array of 8-bit integers
+ * @param hex  A hex string to convert
+ * @returns An array of 8-bit integers
+ */
+ export const hexStringToArray = (hex: string) => {
```
twiss, line 87: This line as well

twiss (approved):

PR #148, opened 2022-08-05
Set `allowInsecureVerificationWithReformattedKeys` in global config

larabr:
TODO:
- [x] point to released OpenPGP.js patch (for fixed TS definition)

### DIFF
--- a/lib/openpgp.d.ts
+++ b/lib/openpgp.d.ts
@@ -1,2 +1,0 @@
-export * from 'openpgp/lightweight';
-export function setConfig(): void;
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -29,7 +29,7 @@
         "@openpgp/asmcrypto.js": "^2.3.2",
         "@openpgp/tweetnacl": "^1.0.3",
         "@openpgp/web-stream-tools": "^0.0.10",
-        "openpgp": "npm:@protontech/openpgp@~5.3.0"
+        "openpgp": "github:openpgpjs/openpgpjs#main"
```
twiss, line 32: Is this required for this change?

twiss (commented):
The main change in this PR looks good, but:

package.json
twiss: Ah sorry, didn't read the PR description. Yeah, if it's not critical, maybe let's wait for that?

twiss (approved):

PR #151, opened 2022-08-08
Fix `removeTrailingSpaces` for \r\n line endings and do not log verification errors

larabr:
TODO
- [x] point to OpenPGP.js v5.4


twiss (commented):
(Seems good to me, pending the TODO)

twiss (approved):

PR #152, opened 2022-08-09
Export `removeTrailingSpaces()` as part of utils

larabr:
To be importable by the apps, as it's needed for implementing fallback verification following the fix in #151


larabr:
Not needed, at least for now

PR #153, opened 2022-08-16
CI: move from CircleCI to Github Actions

larabr:
See commit messages.

I've looked into caching the browsers to avoid downloading them at every run, but their dependencies would still need to be installed, and as of now the overhead of downloading the browsers does not seem that high.


.github/workflows/tests.yml
```diff
@@ -0,0 +1,32 @@
...
+  tests:
+    name: Tests
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v3
+
+      - name: Install dependencies
+        run: npm i
```
dhoko, line 17: you should use `npm ci` rather than `npm i` as it will validate a predictable state.

dhoko: You should provide a lock btw.

PR #157, opened 2022-09-23
Replace MailParser with jsmimeparser in `processMIME`

### LINKED ISSUE #143: Replace or strip down MailParser code
It's a very large library with redundant/unclear dependencies.

larabr:
Fix #143 and use a library specifically built for the browser.
Also, rewrite the `processMIME` module in TS.

Breaking changes in `processMIME` returned data:
- remove `attachment.checksum`, use `attachment.contentId` as unique attachment identifier
- remove `attachment.generatedFileName`, use directly `attachment.fileName`
  - if the attachment did not have a filename set, we still generate one and determine the extension based on the content-type, but we support fewer extensions than before (to keep the library light)
- rename `attachment.length` to `attachment.size`

TODO:
- [ ] point to released jsmimeparser lib


test/message/processMIME.spec.ts
```diff
@@ -78,12 +79,11 @@ describe('processMIME', () => {
...
         const [attachment] = attachments;
         expect(attachment.fileName).to.equal('test.txt');
-        expect(attachment.generatedFileName).to.equal('test.txt');
         expect(attachment.contentType).to.equal('text/plain');
         expect(attachment.contentDisposition).to.equal('attachment');
-        expect(attachment.checksum).to.equal('94ee2b41f2016f2ec79a7b3a2faf920e');
+        expect(attachment.contentId.indexOf('pmcrypto')).to.not.equal(-1);
         expect(attachment.content).to.be.instanceOf(Uint8Array);
-        expect(utf8ArrayToString(attachment.content)).to.equal('this is the attachment text\r\n');
+        expect(utf8ArrayToString(attachment.content)).to.equal('this is the attachment text\n');
```
larabr, line 86: Re: the change in EOL, apparently MailParser automatically made the conversion, but that was a bug, as I checked with their latest release and they also return only `\n` (which is the EOL originally used in the attachment).

lib/message/processMIME.ts
```diff
@@ -0,0 +1,223 @@
...
+import { serverTime } from '../serverTime';
+import { parseMail, generateFileName } from './parseMail';
+import { readSignature, PublicKey, MaybeArray, Signature as OpenPGPSignature } from '../openpgp';
+import { arrayToHexString, utf8ArrayToString } from '../utils';
+
+import type { Attachment } from './parseMail';
+
+export interface MIMEAttachment extends Attachment {
+    /**
+     * `content-id` header value if present, otherwise a unique value derived from the checksum
```
larabr, line 15:
```suggestion
     * `content-id` header value if present, otherwise a unique random value
```

twiss (approved):

PR #158, opened 2022-09-23
Simplify and speed up `utils.arrayToHexString`

larabr:


lib/utils.ts
```diff
@@ -96,18 +96,12 @@ export const hexStringToArray = (hex: string) => {
...
-            h = '0' + h;
-        }
-        r.push('' + h);
+    const res = [];
+    for (let c = 0; c < bytes.length; c++) {
+        const hex = bytes[c].toString(16);
+        res.push(`${hex.length < 2 ? '0' + hex : hex}`);
     }
-    return r.join('');
+    return res.join('');
```
twiss, line 104: :+1: This can even be:
```js
    return Array.from(bytes, byte => byte.toString(16).padStart(2, '0')).join('');
```

larabr: That is optimised in Chrome, but much slower in Safari and especially Firefox

twiss: Hmm, ok

lib/utils.ts
```diff
@@ -96,18 +96,12 @@ export const hexStringToArray = (hex: string) => {
...
-    while (c < e) {
-        h = bytes[c++].toString(16);
-        while (h.length < 2) {
-            h = '0' + h;
-        }
-        r.push('' + h);
+    const res = [];
+    for (let c = 0; c < bytes.length; c++) {
+        const hex = bytes[c].toString(16);
+        res.push(`${hex.length < 2 ? '0' + hex : hex}`);
```
twiss, line 102:
```suggestion
        res.push(hex.length < 2 ? '0' + hex : hex);
```
This can also be simplified (or does it also slow things down? 😅)

twiss (approved):

PR #166, opened 2023-02-20
Add context support (domain separation for signatures)

larabr:
The context is an abstraction we use to add domain separation to OpenPGP signatures.
 It's implemented by adding notation data to signatures, marked as critical, so that the resulting signature can only be verified by a verifier who expects the context to be present.
 
TODO:

- [x]  point to released openpgpjs

### DIFF
--- a/lib/message/context.ts
+++ b/lib/message/context.ts
@@ -1,0 +1,102 @@
+/**
+ * The context is an abstraction we use to add domain separation to OpenPGP signatures.
+ * It's implemented by adding notation data to signatures, which may be marked as critical, so that
+ * the resulting signature can only be verified by a verifier who expects the context to be present.
+ */
+import { SignaturePacket, RawNotation, PartialConfig } from '../openpgp';
+import { normalizeDate } from './utils';
+
+// Different contexts will affect the notation's value, not the name
+const CONTEXT_NOTATION_NAME = 'context@proton.ch';
+
+export interface ContextSigningOptions {
+    /** context identifier */
+    value: string,
+    /**
+     * Whether verification must fail unless the verifier is context-aware:
+     * if `true`, the signature cannot be verified unless the verifier expects the same context.
+     * If `false`, the signature will verify successfully even if the verifier does not expect any context.
+     * This should be set to `true` as long as all clients verifying the signature are able to process contexts.
+     */
+    critical: boolean
+}
+
+export type ContextVerificationOptions = {
+    value: string,
+    /**
+     * Whether the signature must include context info for verification to succeed:
+     * if `true`, a signature with no context won't verify. If `false`, it will.
+     * Note: if the context is not required, but a different context is found in the signature, verification will always fail.
+     */
+    required: boolean,
+    requiredAfter?: undefined
+} | {
+    value: string,
+    required?: undefined
+    /**
+     * Signatures created after this date must include context info for verification to succeed (same as `required: true`).
+     * For signatures created before the given date, the verification behavior is equivalent to `required: false`
+     * */
+    requiredAfter: Date,
+};
+
+/**
+ * Translate a `contextID` string into an OpenPGP notation object, which can be signed as part of the message
+ * @param contextValue - context identifier
+ * @param critical - whether the notation should be critical (if so, verification will fail if in the wrong context,
+ *  i.e. if the caller did not declare the notation as known)
+ */
+export const getNotationForContext = (contextValue: string, critical: boolean): RawNotation => ({
+    name: CONTEXT_NOTATION_NAME,
+    value: new TextEncoder().encode(contextValue),
+    humanReadable: true,
+    critical
+});
+
+/**
+ * Confirm context validity by finding the corresponding signature notation.
+ * NB: signature validity is not checked.
+ */
+export const isValidSignatureContext = (contextOptions: ContextVerificationOptions, signature: SignaturePacket) => {
+    const { value: expectedValue, required, requiredAfter } = contextOptions;
+    const isContextRequired = requiredAfter ? signature.created! >= normalizeDate(requiredAfter) : !!required;
+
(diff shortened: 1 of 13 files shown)


lib/message/context.ts
```diff
@@ -0,0 +1,57 @@
...
+import type { SignaturePacket, RawNotation, PartialConfig } from '../openpgp';
+
+// Different contexts will affect the notation's value, not the name
+const CONTEXT_NOTATION_NAME = 'context@proton.ch';
+
+/**
+ * Translate a `context` string into an OpenPGP notation object, which can be signed as part of the message
+ * @param context - context identifier
+ */
+export const getNotationFromContext = (context: string): RawNotation => ({
```
twiss, line 15: Super nitpicky but this should be called `getNotationForContext` or `createNotationForContext`; `getNotationFromContext` would imply that you're taking the notation out of the context parameter somehow :)

lib/message/context.ts
```diff
@@ -0,0 +1,57 @@
...
+        name === CONTEXT_NOTATION_NAME && new TextDecoder().decode(value) === context
+    ))
+);
+
+/**
+ * Signatures with context data can only be verified with a config that declares the context notation data as known
+ */
+export const getConfigForContextVerification = (config: PartialConfig) => ({
+    ...config,
+    knownNotations: [CONTEXT_NOTATION_NAME] // we can overwrite the field as we currently don't use any other known notations
```
twiss, line 40: We should just remove the default values, and then this comment can be removed

test/message/context.spec.ts
```diff
@@ -0,0 +1,171 @@
+import { expect } from 'chai';
+import { verifyMessage, signMessage, generateKey, readSignature, readMessage, decryptMessage, encryptMessage, readKey } from '../../lib';
+import { VERIFICATION_STATUS } from '../../lib/constants';
+
+// verification without passing context should fail
+// verification passing wrong context should fail
+// verification of unsigned notation data should fail
```
twiss, line 7: Could you also add a test case of verifying a signature that doesn't have any context notation data, to check that it fails to verify (when a context is passed to `verifyMessage`)? I know it's almost the same as the unsigned one, but since it's the main case that we care about, I think it would be nice to have an explicit test for it (and not just one "hidden" as an edge case)

twiss (approved):
Seems good to me (once indeed we use the released OpenPGP.js)

lib/message/verify.js
```diff
@@ -16,16 +17,17 @@ const { NOT_SIGNED, SIGNED_AND_VALID, SIGNED_AND_INVALID } = VERIFICATION_STATUS
  *           verified: Promise<Boolean>,
  *           signature: Promise<openpgp.signature.Signature>
  *         }
+ * @param {String} context - expected signature context
```
larabr, line 20:
```suggestion
 * @param {Object} context - context verification options
```

lib/message/context.ts
```diff
@@ -0,0 +1,102 @@
...
+     * Note: if the context is not required, but a different context is found in the signature, verification will always fail.
+     */
+    required: boolean,
+    requiredFrom?: undefined
+} | {
+    id: string,
+    required?: undefined
+    /**
+     * Signatures created after this date must include context info for verification to succeed (same as `expect: true`).
+     * For signatures created before the given date, the verification behavior is equivalent to `expect: false`
```
twiss, line 38:
```suggestion
     * Signatures created after this date must include context info for verification to succeed (same as `required: true`).
     * For signatures created before the given date, the verification behavior is equivalent to `required: false`
```

lib/message/context.ts
```diff
@@ -0,0 +1,102 @@
...
+     * */
+    requiredFrom: Date,
```
twiss, line 40: I would call this
```suggestion
    requiredAfter: Date,
```

lib/message/context.ts
```diff
@@ -0,0 +1,102 @@
...
+    critical
+});
+
+/**
+ * Confirm context validity by finding the corresponding signature notation.
+ * NB: signature validity is not checked.
+ */
+export const isValidSignatureContext = (contextOptions: ContextVerificationOptions, signature: SignaturePacket) => {
+    const { id: contextID, required, requiredFrom } = contextOptions;
+    const isContextRequired = requiredFrom ? normalizeDate(requiredFrom) <= signature.created! : !!required;
```
twiss, line 62: And then this reads slightly more naturally as
```suggestion
    const { id: contextID, required, requiredAfter } = contextOptions;
    const isContextRequired = requiredAfter ? signature.created! > normalizeDate(requiredAfter) : !!required;
```

lib/message/context.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    // `rawNotations` are always hashed (i.e. signed), otherwise OpenPGP's ignores them on parsing
+    const contextNotations = signature.rawNotations.filter(({ name }) => (name === CONTEXT_NOTATION_NAME));
+    const matchingContext = contextNotations.find(({ value }) => (new TextDecoder().decode(value) === contextID));
+
+    if (matchingContext) {
+        return true;
+    } else if (contextNotations.length > 0) {
+        // mismatching context are present
```
twiss, line 71:
```suggestion
        // mismatching context is present
```

lib/message/context.ts
```diff
@@ -0,0 +1,102 @@
...
+ */
+import { SignaturePacket, RawNotation, PartialConfig } from '../openpgp';
+import { normalizeDate } from './utils';
+
+// Different contexts will affect the notation's value, not the name
+const CONTEXT_NOTATION_NAME = 'context@proton.ch';
+
+export interface ContextSigningOptions {
+    /** context identifier */
+    id: string,
```
twiss, line 14: I would call this `value` rather than `id`, even if it's a bit generic, it really could be anything in principle, not just an ID; e.g. in the case of Mail we could put some headers there, etc.

twiss (approved):

PR #174, opened 2023-04-05
Add build step before publishing the package

lkpch:
Add `prepack` npm script that automatically runs before `npm publish`. The script runs basic compilation TypeScript -> CommonJS and outputs the result to `dist` folder.
This way, it will be possible to import pmcrypto library as CommonJS dependency. It's critical in some contexts, like for example Playwright.


twiss (approved):
:+1: Thanks!

PR #178, opened 2023-06-24
Add `generateForwardingMaterial`

larabr:
Add `generateForwardingMaterial` to generate keys that allow decryption of forwarded messages (https://github.com/ProtonMail/openpgpjs/pull/6).

Supersede and close #90 .


lib/key/forwarding.js
```diff
@@ -0,0 +1,86 @@
...
+    const proxyParameter = dC.modInv(n).mul(dB).mod(n).toUint8Array('le');
+
+    return proxyParameter;
+}
+
+/**
+ * Generate a forwarding key for the final recipient, as well as the corresponding proxy factor.
+ * The key in input must be a v4 primary key and must have at least one ECDH subkey using curve25519 (legacy format)
+ * @param originalKey       ECC primary key of original recipient
+ * @param forwardingUserIds array of user IDs of forwarding key
```
twiss, line 42: `forwarding` is too vague IMO, it should be either `forwardee` or `finalRecipient` (I would go with the former as that's the term used in the draft spec)

lib/key/forwarding.js
```diff
@@ -0,0 +1,86 @@
...
+ * @param subkeyId          (optional) keyid of the ECDH subKey to use for the original recipient
```
twiss, line 43:
```suggestion
 * @param subkeyID          (optional) Key ID of the ECDH subkey to use from the original key
```

lib/key/forwarding.js
```diff
@@ -0,0 +1,86 @@
...
+ * @returns {Promise<Object>}                   The generated key object in the form:
+ *          { proxyFactor: Uint8Array, finalRecipientKey: PrivateKey }
```
twiss, line 45: Also here I would go with the terms from the [draft](https://github.com/wussler/draft-forwarding/blob/main/draft-wussler-openpgp-forwarding.md):
```suggestion
 *          { proxyParameter: Uint8Array, forwardeeKey: PrivateKey }
```
(which for the first one you already did below)

lib/key/forwarding.js
```diff
@@ -0,0 +1,86 @@
...
+    const n = BigInteger.new('0x1000000000000000000000000000000014def9dea2f79cd65812631a5cf5d3ed'); // 2^252 + 0x14def9dea2f79cd65812631a5cf5d3ed
+    const proxyParameter = dC.modInv(n).mul(dB).mod(n).toUint8Array('le');
+
+    return proxyParameter;
+}
+
+/**
+ * Generate a forwarding key for the final recipient, as well as the corresponding proxy factor.
+ * The key in input must be a v4 primary key and must have at least one ECDH subkey using curve25519 (legacy format)
+ * @param originalKey       ECC primary key of original recipient
```
twiss, line 41:
```suggestion
 * @param forwarderKey       ECC primary key of forwarder (original recipient)
```

test/key/forwarding.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+import { enums, KeyID } from '../../lib/openpgp';
+import { generateKey, generateForwardingMaterial, encryptMessage, decryptMessage, readMessage } from '../../lib';
+import { computeProxyParameter } from '../../lib/key/forwarding';
+import { hexStringToArray, concatArrays } from '../../lib/utils';
+
+async function proxyTransform(
+    armoredCiphertext: string,
+    proxyParameter: Uint8Array,
+    originalSubKeyID: KeyID,
+    finalRecipientSubKeyID: KeyID
```
twiss, line 14:
```suggestion
    forwarderSubkeyID: KeyID,
    forwardeeSubkeyID: KeyID
```

twiss:
Also - for future-proofness, I think it'd actually be better to handle fingerprints than key IDs. That way, it'll still work with v6 PKESKs.

lib/key/forwarding.js
larabr: I think Aron wanted to move away from 'forwardee' and similar since it's very confusing to distinguish what is what.
It also shouldn't be 'finalRecipient' since effectively we want this to be the original recipient userID.
The field description makes it clearer what the value is, but maybe I can extend the name to `userIDsForForwardingKey` ?

twiss (approved):

PR #184, opened 2023-09-06
`generateForwardingMaterial`: set forwarding flag on encryption subkeys

larabr:
A specific flag needs to be set for forwarding subkeys instead of the standard EtEr ones. In fact, encrypting to them directly is not allowed.

Also, setting `config.allowForwardedMessages` is now required to decrypt (this was previously not enforced due to a bug in OpenPGP.js)

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
         "@openpgp/tweetnacl": "^1.0.3",
         "@openpgp/web-stream-tools": "^0.0.13",
         "jsmimeparser": "npm:@protontech/jsmimeparser@^3.0.1",
-        "openpgp": "npm:@protontech/openpgp@~5.9.1-0"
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -34,7 +34,7 @@
         "@openpgp/tweetnacl": "^1.0.3",
         "@openpgp/web-stream-tools": "^0.0.13",
         "jsmimeparser": "npm:@protontech/jsmimeparser@^3.0.1",
-        "openpgp": "npm:@protontech/openpgp@~5.9.1-0"
+        "openpgp": "larabr/openpgpjs#forwarding-generate-key-0x40"
```
twiss, line 37: (Once https://github.com/ProtonMail/openpgpjs/pull/8 is merged, could you publish it and keep pointing to npm?)

twiss (approved):

PR #186, opened 2023-10-02
`isForwarding`: add support for encrypted private keys

larabr:
Needed to determine status of inactive address keys.

### DIFF
--- a/lib/key/forwarding.ts
+++ b/lib/key/forwarding.ts
@@ -66,7 +66,7 @@ async function getEncryptionKeysForForwarding(forwarderKey: PrivateKey) {
 /**
  * Whether the given key can be used as input to `generateForwardingMaterial` to setup forwarding.
  */
-export const doesKeySupportForwarding = (forwarderKey: PrivateKey) => (
+export const doesKeySupportForwarding = async (forwarderKey: PrivateKey) => (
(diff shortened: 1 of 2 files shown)


lib/key/forwarding.ts
```diff
@@ -66,13 +67,14 @@ async function getEncryptionKeysForForwarding(forwarderKey: PrivateKey) {
  * Whether the given key can be used as input to `generateForwardingMaterial` to setup forwarding.
  */
 export const doesKeySupportForwarding = (forwarderKey: PrivateKey) => (
-    getEncryptionKeysForForwarding(forwarderKey)
+    forwarderKey.isDecrypted() && getEncryptionKeysForForwarding(forwarderKey)
         .then((keys) => keys.length > 0)
         .catch(() => false)
```
twiss, line 72: This makes the type of the return value into `boolean | Promise<boolean>`, which is probably unintentional. To fix that, you'd have to make the function async. And I would also then just write this as
```suggestion
    try {
      return forwarderKey.isDecrypted() && getEncryptionKeysForForwarding(forwarderKey).length > 0;
    } catch {
      return false;
    }
```

larabr: `getEncryptionKeysForForwarding` is a promise, and adding awaits makes it more verbose than it needs to be IMO

twiss (approved):

PR #190, opened 2023-11-23
Upgrade to OpenPGP.js v6

larabr:
Non-breaking changes:
- added support for RFC9580 (new OpenPGP spec), including:
  - reading and generating v6 packet versions
  - processing and generating Curve448 keys, and Curve25519 in new format
  - ...

Breaking changes:
- **for downstream TS compiler,** using the new "node16" or "Bundler" module resolution is required, as e.g. the noble-hashes lib uses `exports` only (if this is too problematic, we could temporarily add legacy entrypoints to that dependency, since we currently use a fork).
- limited support for (legacy, non-standardized) AEAD-encrypted v4 keys as per draft4880bis
- `checkKeyCompatibility` will reject v5 keys
- enum name changes related to legacy curve25519, eddsa, and nist curves
- ...

TODO:
- [x] rebase on top of TS fix for lightweight build (to be released in next openpgpjs v6 pre-release)
- [x] release protontech/v6 and point to it
- [x] test integration with web-clients, esp. TS config
- post-canary:
  - [x]  we currently reject keys in new eddsa/xecdh format (in `checkKeyStrength`) (allow for v6 keys only)
  - [x] accept sha3 in generated key prefs (v6 keys only)
  - [ ] make explicit `date` inputs mandatory?
- [ ] ...


lib/key/forwarding.ts
```diff
@@ -1,29 +1,16 @@
...
-        BigIntegerInterface.setImplementation(FallbackBigInteger);
-    }
+    // noble-hashes's BigInteger export automatically imports BN.js (as BigInt fallback).
+    // Because of its large size, and the fact that it's not always needed, openpgpjs v6 only imports noble-hashes dynamically,
+    // so we do the same here.
+    if (MaybeUninitializedBigInteger) return MaybeUninitializedBigInteger;
 
-    return BigIntegerInterface;
+    MaybeUninitializedBigInteger = await import('@openpgp/noble-hashes/biginteger').then((mod) => mod.default);
+    return MaybeUninitializedBigInteger;
```
larabr, line 13: Not sure about this, also in the context of the chunk loading issues.
Perhaps it's best to introduce a BigIntegerInterface module copy here, or expose the one in OpenPGP.js, so that we (1) avoid the shared state with noble-hashes and (2) can only dynamically import BN.js .

PR #191, opened 2023-11-28
Add `checkKeyCompatibility()`

### LINKED ISSUE #122: Replace `keyCheck` and simplify `keyInfo`
- `keyCheck` should be removed since it is not used in our apps, and it is strongly coupled with `keyInfo`. The checks it does should be re-implemented in a different function (e.g. named `checkPersonalKey`), taking an `OpenPGPKey` in input, instead of a custom object. It should check key strength requirements through the function(s) implemented in #121 , as well as binding signature validity, whether the key can encrypt etc (same checks as API).
- `keyInfo` should be removed, ~or simplified to not include key checks. It is currently used in apps [to collect minimal key metadata](https://github.com/ProtonMail/WebClients/blob/8d7f8a902034405988bd70431c714e9fdbb37a1d/applications/mail/src/app/helpers/message/messageAttachPublicKey.ts#L41).~

larabr:
To check whether a public key is compatible with all clients.

Also, drop outdated `keyInfo` and `keyCheck` functions (close #122).


test/key/check.spec.ts
```diff
@@ -80,26 +78,47 @@ describe('key checks', () => {
...
+Gw0vQaiZn6HGITQw5nBGvXQPF9VpFpsXV9x/08dIdfZLAQVdQowgeBsxCw==
+=JIkN
+-----END PGP PUBLIC KEY BLOCK-----` });
+        expect(
+            () => checkKeyCompatibility(key)
+        ).to.throw(/key algorithm is currently not supported/);
+    });
 
+    it('compatibility - it rejects a v6 key', async () => {
+        // currently reading the key fails, but once OpenPGP.js v6 is integrated, we'll can test `checkKeyCompatibility`
```
twiss, line 100:
```suggestion
        // currently reading the key fails, but once OpenPGP.js v6 is integrated, we can test `checkKeyCompatibility`
```

twiss (approved):

PR #195, opened 2024-02-20
Tests: load stream polyfills if needed (fix Safari 13), add README note about polyfilling requirement

larabr:
Since v7.2.2, the web-stream-tools version used by pmcrypto diverges from the one of OpenPGP.js, and it does not automatically polyfill streams.
Without polyfilling, some of the lib functions do not work on legacy browsers when streamed inputs are used. Affected functions include those that internally process streams (independently of OpenPGP.js), such as `encryptMessage`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,9 +98,9 @@ const { data: decryptedData, verified } = await decryptMessage({
 to encrypt (and/or sign), pass the stream to `textData` or `binaryData` based on the streamed data type. Similarly, to decrypt and verify, the input options are the same as the non-streaming case. However, if `armoredMessage` (or `binaryMessage`) is a stream, the decryption result needs to be handled differently:
 ```js
 // explicitly loading stream polyfills for legacy browsers is required since v7.2.2
-  if (!globalThis.TransformStream) {
-      await import('web-streams-polyfill/es6');
-  }
+if (!globalThis.TransformStream) {
+  await import('web-streams-polyfill/es6');
+}
 
 const { data: dataStream, verified: verifiedPromise } = await decryptMessage({
   message: await readMessage({ armoredMessage: streamedArmoredMessage }),
--- a/README.md
+++ b/README.md
@@ -97,6 +97,11 @@ const { data: decryptedData, verified } = await decryptMessage({
 **For streamed inputs:**
 to encrypt (and/or sign), pass the stream to `textData` or `binaryData` based on the streamed data type. Similarly, to decrypt and verify, the input options are the same as the non-streaming case. However, if `armoredMessage` (or `binaryMessage`) is a stream, the decryption result needs to be handled differently:
 ```js
+// explicitly loading stream polyfills for legacy browsers is required since v7.2.2
+  if (!globalThis.TransformStream) {
+      await import('web-streams-polyfill/es6');
+  }
+
 const { data: dataStream, verified: verifiedPromise } = await decryptMessage({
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -97,6 +97,13 @@ const { data: decryptedData, verified } = await decryptMessage({
 **For streamed inputs:**
 to encrypt (and/or sign), pass the stream to `textData` or `binaryData` based on the streamed data type. Similarly, to decrypt and verify, the input options are the same as the non-streaming case. However, if `armoredMessage` (or `binaryMessage`) is a stream, the decryption result needs to be handled differently:
 ```js
+// explicitly loading stream polyfills for legacy browsers is required since v7.2.2
+if (!globalThis.TransformStream || !globalThis.WritableStream) {
+  const ponyfilledStreams = await import('web-streams-polyfill/es6');
```
twiss, line 102: The polyfill already overwrites the global variables. If you want to do so manually (which I think is reasonable for clarity), you need the ponyfill.
```suggestion
  const ponyfilledStreams = await import('web-streams-polyfill/ponyfill/es6');
```

README.md
```diff
@@ -97,6 +97,13 @@ const { data: decryptedData, verified } = await decryptMessage({
...
+  globalThis.TransformStream = globalThis.TransformStream || ponyfilledStreams.TransformStream;
+  globalThis.WritableStream = globalThis.WritableStream || ponyfilledStreams.WritableStream;
```
twiss, line 104: The way web-steam-tools used to do it is, if one of them is missing, overwrite all of them. And I think that's safer, to avoid unexpected incompatibilities etc. So probably we should do the same here (and in the web apps, if we aren't already):
```suggestion
  ({ ReadableStream, WritableStream, TransformStream } = ponyfilledStreams);
```

test/message/encryptMessage.spec.ts
```diff
@@ -10,17 +10,28 @@ import { testPrivateKeyLegacy } from './decryptMessageLegacy.data';
...
+    const { WritableStream, TransformStream } = globalThis;
+
+    before('downgrade openpgp config and load stream polyfills', async () => {
         globalConfig.minRSABits = 512;
+
+        // load stream ponyfills if needed
+        const ponyfilledStreams = await import('web-streams-polyfill/es6');
+        globalThis.TransformStream = TransformStream || ponyfilledStreams.TransformStream;
+        // @ts-ignore
+        globalThis.WritableStream = WritableStream || ponyfilledStreams.WritableStream;
```
twiss, line 28: Same here

README.md
```diff
@@ -97,6 +97,11 @@ const { data: decryptedData, verified } = await decryptMessage({
 **For streamed inputs:**
 to encrypt (and/or sign), pass the stream to `textData` or `binaryData` based on the streamed data type. Similarly, to decrypt and verify, the input options are the same as the non-streaming case. However, if `armoredMessage` (or `binaryMessage`) is a stream, the decryption result needs to be handled differently:
 ```js
+// explicitly loading stream polyfills for legacy browsers is required since v7.2.2
+  if (!globalThis.TransformStream) {
+      await import('web-streams-polyfill/es6');
+  }
```
twiss, line 103:
```suggestion
if (!globalThis.TransformStream) {
  await import('web-streams-polyfill/es6');
}
```

twiss (approved):

PR #199, opened 2024-04-11
`argon2`: reload wasm module after timeout when using memory heavy params

larabr:
We manually reload the module if no memory-heavy (128MB+) argon2 computation has been requested in a while, to deallocate the memory.
This is better than reloading the module every time (automatically done if the memory exceeds `Argon2S2K.ARGON2_WASM_MEMORY_THRESHOLD_RELOAD`), as doing so comes with a performance hit.

### DIFF
--- a/lib/crypto/argon2.ts
+++ b/lib/crypto/argon2.ts
@@ -12,8 +12,36 @@ export interface Argon2Options {
     params: Argon2Params
 }
 
+// We manually reload the module if no memory-heavy (128MB+) argon2 computation has been requested in a while,
+// to deallocate the memory.
+// This is better than reloading the module every time (automatically done if the memory exceeds `Argon2S2K.ARGON2_WASM_MEMORY_THRESHOLD_RELOAD`),
+// as doing so comes with a performance hit.
+const SECOND = 1000;
+const TimeoutHandler = {
+    id: undefined,
+    cancelReloadingTimeout: (memoryExponent: number) => (
+        memoryExponent > ARGON2_PARAMS.MINIMUM.memoryExponent && clearTimeout(TimeoutHandler.id)
+    ),
+    setupReloadingTimeout: (memoryExponent: number) => {
+        const shouldReloadAfterTimeout = memoryExponent > ARGON2_PARAMS.MINIMUM.memoryExponent &&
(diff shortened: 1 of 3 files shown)


twiss (approved):

lib/crypto/argon2.ts
```diff
@@ -12,8 +12,36 @@ export interface Argon2Options {
...
+// as doing so comes with a performance hit.
+const SECOND = 1000;
+const TimeoutHandler = {
+    id: undefined,
+    cancelReloadingTimeout: (memoryExponent: number) => (
+        memoryExponent > ARGON2_PARAMS.MINIMUM.memoryExponent && clearTimeout(TimeoutHandler.id)
+    ),
+    setupReloadingTimeout: (memoryExponent: number) => {
+        const shouldReloadAfterTimeout = memoryExponent > ARGON2_PARAMS.MINIMUM.memoryExponent &&
+            memoryExponent < Argon2S2K.ARGON2_WASM_MEMORY_THRESHOLD_RELOAD;
```
twiss, line 27:
```suggestion
            memoryExponent <= Argon2S2K.ARGON2_WASM_MEMORY_THRESHOLD_RELOAD;
```

lib/crypto/argon2.ts
```diff
@@ -12,8 +12,36 @@ export interface Argon2Options {
...
 
+// We manually reload the module if no memory-heavy (128MB+) argon2 computation has been requested in a while,
+// to deallocate the memory.
+// This is better than reloading the module every time (automatically done if the memory exceeds `Argon2S2K.ARGON2_WASM_MEMORY_THRESHOLD_RELOAD`),
+// as doing so comes with a performance hit.
+const SECOND = 1000;
+const TimeoutHandler = {
+    id: undefined,
+    cancelReloadingTimeout: (memoryExponent: number) => (
+        memoryExponent > ARGON2_PARAMS.MINIMUM.memoryExponent && clearTimeout(TimeoutHandler.id)
```
twiss, line 23: Bit nitpicky, but if we continue to use it, even with a lower memory exponent, it might still be worth it to keep the instance around?

larabr: There is a risk that the memory is never deallocated otherwise, so this is to avoid that possibility

twiss: Ah, right

PR #206, opened 2025-01-24
Fix `isForwardingKey` to support keys without valid encryption subkeys

larabr:


lib/key/forwarding.ts
```diff
@@ -53,15 +55,33 @@ export async function doesKeySupportForwarding(forwarderKey: PrivateKey, date: D
...
+
+    const allDecryptionKeys = await keyToCheck
+        .getDecryptionKeys(undefined, date, undefined, {
+            ...defaultConfig,
+            allowForwardedMessages: true
+        })
+        .catch(() => []); // throws if no valid decryption keys are found
+
+    const hasForwardingKeyFlag = (maybeForwardingSubkey: Subkey) => {
+        const flags = maybeForwardingSubkey.bindingSignatures[0].keyFlags?.[0];
```
twiss, line 74: In theory, it's possible for a key to have multiple binding signatures, and `getDecryptionKeys` will return the key if at least one of them is valid. So, we need to either check that this signature is the valid one, or (probably simpler) check that there's only one binding signature, or that all of them have the required flag.

twiss (approved):

PR #208, opened 2025-02-05
md5: use noble-hashes, and import code dynamically

larabr:
Refactor md5 code as done in OpenPGP.js , and fetch the code only when the function is called.


paulmillr:
Love it!

twiss (approved):

PR #210, opened 2025-02-28
Tests: migrate from karma (deprecated) to web-test-runner

### LINKED ISSUE #135: Tests: incorrect line references reported on error
In tests, the line references given on error are off, due to two separate issues:
- since webpack does not emit source map files for TS, a custom plugin should be set: https://github.com/ryanclark/karma-webpack/issues/109#issuecomment-224961264
- even with the above fix, karma-webpack v5 has a bug that breaks the source map referencing: https://github.com/ryanclark/karma-webpack/issues/493. The proposed temporary solution (set `splitChunks: false`) does not work for us, because chunking is needed in the tests to load the mailparses and openpgp.js. Their next release hopefully fixed the problem.

larabr:
Karma has been deprecated for some time, and the wtr playwright integration makes it easier to run e.g. specific browser versions.
Since wtr does not integrate with webpack, we also switch to esbuild and rollup for the test bundling.

Also, fix #135 .


twiss (approved):
In case you're planning to rebase/merge rather than squash: there's a typo/reduplicated "scalar clamping" in the first commit message. Other than that all LGTM.

PR #223, opened 2025-08-13
TS: declare `Uint8Array<ArrayBuffer>` over generic `Uint8Array`

larabr:
Also, enforce behavior via ESLint.

This is to limit the need to downcast Uint8Array in output when using them with e.g. WebCrypto and Blobs,
following an incompatible interface change between ArrayBuffer and SharedArrayBuffer (enforced in TS v5.9).

More context:
a `Uint8Array` type definition without generic-type indicator defaults to `Uint8Array<ArrayBufferLike>`.
The core of the issue is that in ES2024, SharedArrayBuffer and ArrayBuffer are no longer compatible, hence an ArrayBufferLike (which is also supertype of SharedArrayBuffer) cannot be passed if an ArrayBuffer is expected. This breaks Uint8Array usages which expect an ArrayBuffer (which is the case for e.g. the WebCrypto API).


mmso (approved):
Looks good to me!

